PR #40, opened 2026-09-04
Paginatable: paginate Arrays and other Enumerables, not only relations

VSN2015:
## Summary

`paginated` / `pagination_meta` required an ActiveRecord relation — anything else blew up on `.limit`. Results that never touch the database (an external API response, a search-service hit list, a loaded association, a hand-built list of Structs) had no way to get the same slicing, `X-*` headers and `pagination_meta`.

This PR makes `paginated(collection)` accept **any in-memory Enumerable** while keeping the relation path byte-for-byte unchanged:

- **Relation detection is duck-typed**: anything answering `limit` and `offset` (Relation, association `CollectionProxy`, model class) takes the SQL path exactly as before.
- **Any other non-`Hash` `Enumerable`** (`Array`, `Set`, `Range`, `Enumerator`, …) is materialized **once** via `to_a` (an Enumerator is not consumed twice for count + slice), the total is its size, and the current page comes back as an `Array` — `[]` past the last page. Headers and memoized meta are identical to the relation path.
- **A `Hash` is rejected** with an `ArgumentError` pointing at `.to_a` rather than silently paginated as `[key, value]` pairs; `nil`, String, Integer raise the same error naming the class received.
- `pagination_meta(collection)` accepts the same inputs.
- Extracted `LABEL` for error messages (Cacheable/Deprecatable precedent).

```ruby
def search
  hits = ExternalCatalog.search(params[:q])          # a plain Array
  render_success(data: paginated(hits), meta: pagination_meta)
end
```

## Docs

- `README.md` — Paginatable section gains an "Arrays and other Enumerables work too" paragraph + snippet.
- `docs/concerns/paginatable.md` — intro, "When to use it", `paginated` / `pagination_meta` method docs (the latter was undocumented), a new in-memory example, two Notes bullets.

```
### Added
- **Controllers::Paginatable**: `paginated` / `pagination_meta` accept in-memory
  collections — an `Array`, `Set`, `Range` or any other non-Hash `Enumerable` —
  not only ActiveRecord relations. The collection is materialized once, the
  total is its size and the current page comes back as an `Array` with the same
  `X-Total-Count` / `X-Page` / `X-Per-Page` / `X-Total-Pages` headers and
  memoized meta. Relations still paginate in SQL. A `Hash`, `nil` or a
  non-collection raises `ArgumentError` naming the class received.
```

## Test plan

- [x] 13 new examples in `spec/concerns/controllers/paginatable_spec.rb`: default slicing + headers, `page`/`per_page`, past-the-end → `[]`, empty Array, source not mutated, `max_per_page` cap, Range/Set/Enumerator, Enumerator single pass, Array of AR records, meta memoization + fresh meta, Hash/nil/String/Integer rejections, relation still emits `LIMIT`.
- [x] Full suite: 1271 examples, 0 failures (98.32% line coverage).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/paginatable.rb
+++ b/lib/concerns_on_rails/controllers/paginatable.rb
@@ -14,9 +14,19 @@ module ConcernsOnRails
     #       render json: paginated(Article.all)
     #     end
     #   end
+    #
+    # `paginated` also takes an in-memory collection — an Array, Set, Range or
+    # any other non-Hash Enumerable — so results assembled outside the
+    # database (an external API, a loaded association, a hand-built list of
+    # Structs) get the same slicing, headers and `pagination_meta`:
+    #
+    #   def search
+    #     render json: paginated(ExternalCatalog.search(params[:q]))
+    #   end
     module Paginatable
       extend ActiveSupport::Concern
 
+      LABEL = "ConcernsOnRails::Controllers::Paginatable".freeze
       DEFAULT_PER_PAGE = 25
       DEFAULT_MAX_PER_PAGE = 200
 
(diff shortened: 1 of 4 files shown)


VSN2015:
**Merge order note (from the enhancement loop):** a dry-run merge of all 22 enhancement branches found only 3 conflicting pairs, all resolved by stacking. This PR is the base of one stack:

- **#40** (this) → **#45** Link headers → **#59** signed cursors

#45 is now based on this branch and #59 on #45, so each diff shows only its own change. Merge them in that order (GitHub retargets the next one to `master` automatically when this branch is deleted on merge). Every other enhancement PR is independent and merges cleanly in any order — README sections never overlap.

VSN2015:
**Stack update:** #62 (Paginatable `total:` for pre-paginated pages) is also based on #45, so the pagination stack is now **#40 → #45 → { #59, #62 }** — merge #40, then #45, then #59 and #62 in either order.

VSN2015 (commented):
LGTM! Extending Paginatable to accept non-relation Enumerables (Arrays, Sets, Ranges) brings consistent pagination headers and metadata to in-memory collections and external API results.

lib/concerns_on_rails/controllers/paginatable.rb
```diff
@@ -35,40 +45,49 @@ def paginate_by(per_page: DEFAULT_PER_PAGE, max_per_page: DEFAULT_MAX_PER_PAGE)
         end
       end
 
-      # Apply pagination to a relation and set the standard response headers.
-      # Returns the paginated relation; the metadata is memoized so a follow-up
-      # `pagination_meta` (no argument) reuses it. Safe on empty relations.
-      def paginated(relation)
+      # Apply pagination to a relation or an in-memory collection and set the
+      # standard response headers. A relation comes back as a relation with
+      # LIMIT/OFFSET applied (still lazy); an Enumerable comes back as the
```
VSN2015, line 50: Materializing non-relation Enumerables once into an Array avoids multiple passes or consuming one-way Enumerators twice during count and slice.

lib/concerns_on_rails/controllers/paginatable.rb
```diff
@@ -80,13 +99,31 @@ def pagination_meta(relation = nil)
...
-      # offset are irrelevant, a custom SELECT list would turn into the invalid
-      # COUNT(a, b), and count(:all) keeps DISTINCT semantics. A grouped
-      # relation counts as a Hash (group => count); the meaningful total is the
-      # number of groups.
-      def paginatable_total(relation)
-        counted = relation.except(:order, :limit, :offset, :select).count(:all)
+      # Relations — anything answering `limit` and `offset`: an
+      # ActiveRecord::Relation, an association CollectionProxy, a model class —
+      # pass through untouched so they keep paginating in SQL. Any other
+      # non-Hash Enumerable is materialized ONCE into an Array, so an
```
VSN2015, line 105: Gracefully distinguishing between ActiveRecord relations (`limit`/`offset`) and plain Arrays keeps database queries optimized in SQL while supporting in-memory lists.

PR #41, opened 2026-09-04
SoftDeletable: restore_all / really_destroy_all keep caller predicates on the soft-delete column

VSN2015:
## Summary

**Bug fix.** `restore_all` routed through the `soft_deleted` scope and `really_destroy_all` through a bare `unscope(where: deleted_at)`. Both peel the default scope's `deleted_at IS NULL` off — and, with it, **every predicate the caller put on that column**:

```ruby
User.deleted_within(1.hour).restore_all      # restored the WHOLE trash can
User.where(deleted_at: from..to).restore_all # same
User.only_deleted.really_destroy_all         # widened to the whole relation (documented as a gotcha)
```

This is the same defect `publish_all` fixed in 1.27, flagged then as a SoftDeletable follow-up.

## Fix

A private `soft_delete_without_default_scope` peels **only the default scope's own predicate**: unscope the column, then put back the predicates the caller added on it. "Added by the caller" is the relation's where clause minus `default_scoped`'s — the same structural `WhereClause` arithmetic Rails' own `merge`/`except` rely on. Consequences:

- Predicates on **other** columns are never touched — including a host model's own `default_scope { where(tenant_id: …) }`, which survives.
- With `default_scope: false` nothing is peeled.
- A bare `User.restore_all` on a default-scoped model still restores the whole trash can (primary use, unchanged).
- Fast path (single `UPDATE`) and per-record path both build on the same relation.
- `only_deleted.really_destroy_all` now purges the trash and nothing else; `deleted_within(30.days).really_destroy_all` purges recent trash. `User.where(author: x).really_destroy_all` still hard-deletes x's rows soft-deleted or not.

The **scopes** themselves still `unscope` the column (unchanged, and used inside the default scope itself), so the docs now state the chaining rule: `soft_deleted.where(...)`, not `where(...).soft_deleted`.

## Docs

- `README.md` — bulk-operations block gains the three "undo a bulk delete / purge recent trash / empty the trash" one-liners; the `really_destroy_all` gotcha paragraph is replaced by the new semantics + scopes-first rule.
- `docs/concerns/soft-deletable.md` — `really_destroy_all` / `restore_all` rows rewritten; new first example "Undo a bulk delete without emptying the whole trash can".

```
### Fixed
- **Models::SoftDeletable**: `restore_all` and `really_destroy_all` now honour a
  caller's predicate on the soft-delete column. Both used to `unscope` the
  column outright to peel off the default scope's `deleted_at IS NULL`, which
  also dropped `deleted_within(1.hour)` / `where(deleted_at: range)` /
  `only_deleted` — so `User.deleted_within(1.hour).restore_all` restored the
  whole trash can and `only_deleted.really_destroy_all` widened to the whole
  relation. Only the default scope's own predicate is peeled now; predicates on
  other columns (a host model's own `default_scope` included) are untouched. The
  scopes themselves still unscope the column: chain `soft_deleted.where(...)`,
  not `where(...).soft_deleted`.
```

## Test plan

- [x] 18 new examples in `spec/concerns/models/soft_deletable_spec.rb`: fast + per-record paths × (`deleted_within`, `where(deleted_at: range)`, `soft_deleted.where(...)`, other-column predicate, bare `restore_all`, empty narrowed relation), single-`UPDATE` fast path with the caller predicate in the SQL, `default_scope: false`, host default scope preserved, three `really_destroy_all` cases. The 10 bug-path examples fail on master.
- [x] Full suite: 1275 examples, 0 failures (98.32%).
- [x] RuboCop clean.


VSN2015:
**Merge order note (from the enhancement loop):** this PR is the base of the second small stack:

- **#41** (this) → **#52** SoftDeletable cascade

#52 is based on this branch (the two only overlapped in `soft_deletable_spec.rb`; the lib changes compose cleanly). Merge #41 first, then #52. All other enhancement PRs are independent of both.

lib/concerns_on_rails/models/soft_deletable.rb
```diff
@@ -82,22 +82,27 @@ def destroy_all
...
-        # soft-deleted rows (only the soft-delete column's predicates are
-        # peeled off). Note that `unscope` also drops a caller's own condition
-        # on that column, so `only_deleted.really_destroy_all` widens to the
-        # whole relation — use `soft_deleted.delete_all` to purge trash only.
+        # Hard-delete every record matching the CURRENT relation — soft-deleted
+        # rows included. Only the default scope's own `deleted_at IS NULL` is
+        # peeled off; a caller's predicate on the column survives, so
+        # `only_deleted.really_destroy_all` purges the trash and nothing else
+        # and `deleted_within(30.days).really_destroy_all` purges recent trash.
         # (Before 1.22 this ignored the relation entirely and hard-deleted the
```
VSN2015, line 90: Selectively removing default scope clauses while retaining user-supplied where predicates prevents accidental mass-restoration or destruction of unintended records.

lib/concerns_on_rails/models/soft_deletable.rb
```diff
@@ -110,6 +115,27 @@ def restore_all
 
         private
 
+        # The current relation with the DEFAULT SCOPE's soft-delete predicate
+        # peeled off — and nothing else. `unscope(where: field)` (what the
+        # scopes do) strips every predicate on the column, the caller's
```
VSN2015, line 120: Preserving caller conditions on `restore_all` fixes chaining with time-bounded scopes like `deleted_within`.

VSN2015 (commented):
LGTM! Great bug fix ensuring that `restore_all` and `really_destroy_all` preserve caller-specified predicates on the soft-delete column (`Article.deleted_within(7.days).restore_all`).

PR #45, opened 2026-09-04
Paginatable + CursorPaginatable: RFC 8288 Link headers (first/prev/next/last)

VSN2015:
## Summary

Clients of both paginators had to assemble the next-page URL themselves from `X-Page`/`X-Total-Pages` or `X-Next-Cursor`. Every paginated response now also carries an **RFC 8288 `Link` header** — the GitHub convention — with URLs rebuilt from the current request (`base_url` + `path`, every other query param preserved):

| Concern | rels |
|---|---|
| Paginatable | `first`, `prev`, `next`, `last` (`page=N`). `prev`/`next` only when such a page exists; past the end `prev` points at the last page; nothing for an empty collection. |
| CursorPaginatable | `next` (`cursor=<X-Next-Cursor>` while more pages exist), `prev` in bidirectional mode, `first` (cursor dropped) once a cursor is in play. `per_page` and the order preset are preserved. |

```
Link: <http://api.example.com/articles?page=1&per_page=10&q=abc>; rel="first",
      <http://api.example.com/articles?page=1&per_page=10&q=abc>; rel="prev",
      <http://api.example.com/articles?page=3&per_page=10&q=abc>; rel="next",
      <http://api.example.com/articles?page=5&per_page=10&q=abc>; rel="last"
```

**Design**

- New **`Support::LinkHeader`** (autoloaded, shared): `available?(controller)` — a request exposing `base_url`/`path`/`query_parameters` (the dependency-free `FakeController` has none, so emission is silently skipped); `url_for(request, drop:, **overrides)` — Rack nested-query encoding, nested params survive, nil/`drop:` removes a key; `append(response, rels)` — **never clobbers** an existing `Link` (Deprecatable's `rel="deprecation"`, CDN hints), appends after it.
- Both macros gain **`link_header: true`**; `paginate_by link_header: false` / `cursor_paginate_by link_header: false` opt out. The `X-*` headers are unchanged either way.
- Behaviour change to be aware of: on by default, so every non-empty paginated response grows by one header. Behind a proxy the URLs name whatever host Rails sees — the docs note `trusted_proxies`/forwarded headers.

**Merge note:** this PR and #40 (Paginatable arrays) both touch `paginatable.rb` (header comment, `paginate_by`, `paginated`). Whichever merges second needs a trivial rebase — I'll do it in the loop as soon as one lands.

## Docs

- `README.md` — both "Response headers" lines.
- `docs/concerns/paginatable.md` — `link_header` config row, `Link` header row, two Notes bullets (proxy/host, append semantics).
- `docs/concerns/cursor-paginatable.md` — `link_header:` config row, `Link` header row, Notes bullet.

```
### Added
- **Controllers::Paginatable / CursorPaginatable**: RFC 8288 `Link` response
  header — `first`/`prev`/`next`/`last` page URLs for Paginatable, `next`
  (+ `prev` in bidirectional mode, `first` once a cursor is in play) for
  CursorPaginatable — rebuilt from the current request with every other query
  param preserved and appended to any existing `Link` header. Opt out with
  `link_header: false` on either macro. Backed by the new shared
  `Support::LinkHeader`.
```

## Test plan

- [x] `spec/concerns/support/link_header_spec.rb` (+8): URL rebuild with ordering/encoding, nil/`drop:` removal, adding a key, nested params, RFC 8288 serialization, append-to-existing, all-nil no-op, `available?` matrix.
- [x] Paginatable (+7, real ActionController via `IntegrationHarness`): full rel set on a middle page with other params preserved, first/last-page omissions, past-the-end `prev`, empty collection → no header, `link_header: false`, appended after a pre-set `Link`, bare-harness skip.
- [x] CursorPaginatable (+5): `next` with the `X-Next-Cursor` token on page 1, `first` once a cursor is in play and no `next` on the last page, `prev` in bidirectional mode with an order preset preserved, `link_header: false`, bare-harness skip.
- [x] Full suite: 1277 examples, 0 failures (98.34%). RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/paginatable.rb
+++ b/lib/concerns_on_rails/controllers/paginatable.rb
@@ -1,5 +1,6 @@
 require "active_support/concern"
 require "concerns_on_rails/support/scalar_param"
+require "concerns_on_rails/support/link_header"
 
 module ConcernsOnRails
   module Controllers
@@ -23,6 +24,12 @@ module ConcernsOnRails
     #   def search
     #     render json: paginated(ExternalCatalog.search(params[:q]))
     #   end
+    #
+    # Every paginated response also carries an RFC 8288 `Link` header with
+    # first/prev/next/last URLs rebuilt from the current request (other query
+    # params preserved) — the GitHub convention, so clients can follow links
+    # instead of computing page numbers. `paginate_by link_header: false` turns
+    # it off.
     module Paginatable
       extend ActiveSupport::Concern
 
@@ -33,15 +40,18 @@ module ConcernsOnRails
       included do
         class_attribute :paginatable_per_page, default: DEFAULT_PER_PAGE
         class_attribute :paginatable_max_per_page, default: DEFAULT_MAX_PER_PAGE
+        class_attribute :paginatable_link_header, default: true
       end
 
       class_methods do
-        # Configure the default page size and the hard cap on per_page.
+        # Configure the default page size, the hard cap on per_page, and whether
+        # the RFC 8288 Link header is emitted.
         # Example:
-        #   paginate_by per_page: 50, max_per_page: 500
-        def paginate_by(per_page: DEFAULT_PER_PAGE, max_per_page: DEFAULT_MAX_PER_PAGE)
+        #   paginate_by per_page: 50, max_per_page: 500, link_header: false
+        def paginate_by(per_page: DEFAULT_PER_PAGE, max_per_page: DEFAULT_MAX_PER_PAGE, link_header: true)
           self.paginatable_per_page = per_page.to_i
           self.paginatable_max_per_page = max_per_page.to_i
+          self.paginatable_link_header = link_header ? true : false
         end
       end
 
@@ -70,6 +80,7 @@ module ConcernsOnRails
 
         @paginatable_meta = { total: total, page: page, per_page: per_page, total_pages: total_pages }
         set_pagination_headers(**@paginatable_meta)
+        set_pagination_links(page: page, total_pages: total_pages)
         records
       end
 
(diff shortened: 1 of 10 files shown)


lib/concerns_on_rails/support/link_header.rb
```diff
@@ -0,0 +1,45 @@
+require "rack/utils"
+
+module ConcernsOnRails
+  module Support
+    # RFC 8288 `Link` response header for the paginators: rebuilds the current
+    # request URL with a few query params changed (`page=3`, `cursor=…`) and
+    # appends `<url>; rel="next"` entries to the response — never clobbering a
+    # Link header something else already set (Deprecatable's rel="deprecation",
+    # CDN preload hints). Shared by Paginatable and CursorPaginatable.
+    module LinkHeader
```
VSN2015, line 10: Encapsulating RFC 8288 link formatting in `LinkHeader` keeps pagination controllers clean and DRY.

lib/concerns_on_rails/controllers/paginatable.rb
```diff
@@ -148,6 +159,30 @@ def set_pagination_headers(total:, page:, per_page:, total_pages:)
         response.set_header("X-Per-Page", per_page.to_s)
         response.set_header("X-Total-Pages", total_pages.to_s)
       end
+
+      # Link: <…?page=1>; rel="first", <…?page=1>; rel="prev", <…?page=3>;
+      # rel="next", <…?page=5>; rel="last". prev/next only when such a page
+      # exists; past the end, prev points at the last page. Nothing is emitted
```
VSN2015, line 165: Rebuilding clean URLs with only relevant pagination parameter overrides preserves other query filters in the Link headers.

VSN2015 (commented):
LGTM! Standard RFC 8288 `Link` headers across both offset and cursor pagination models align API responses with GitHub/REST conventions and simplify client-side traversal.

PR #46, opened 2026-09-04
Filterable: comparison operators (suffix + bracket form) and type coercion

VSN2015:
## Summary

Direct-where filters could only do equality; anything else — a price range, a status list, "not archived", "discontinued IS NULL", a title search — meant a `with:` lambda per field. Filters can now **opt into operators**:

```ruby
filter_by :price, :stock, :status, :name, operators: true      # all of them
filter_by :created_at, operators: %i[gte lte]                  # a subset
filter_by :min_stock, type: :integer, with: ->(rel, v) { … }   # lambda receives Integer 5, not "5"
```

Accepted as a suffix or in bracket form:

| Operator | Param | SQL |
|---|---|---|
| `not` | `?status_not=draft` | `status != 'draft'` |
| `gt` `gte` `lt` `lte` | `?price_gte=10` / `?price[gte]=10&price[lte]=50` | `price >= 10` (cast through the column type) |
| `in` `not_in` | `?status_in=a,b` or `?status_in[]=a` | `status IN ('a','b')` |
| `null` | `?deleted_at_null=true` | `deleted_at IS NULL` (`false` → `IS NOT NULL`) |
| `contains` `starts_with` | `?title_contains=rails` | `title LIKE '%rails%' ESCAPE '\'` (wildcards escaped; ILIKE on PostgreSQL) |

**Design points**

- **Casting follows ActiveRecord.** Comparison values go through the column's own attribute type (so `"10"` compares as a decimal against a decimal column and SQLite's type affinity cannot bite); `type:` (any ActiveModel type name) overrides that and also **pre-casts the value handed to a `with:` lambda**. Equality/`not`/`in`/`null` go through `where`, which casts on its own.
- **Opt-in per filter, direct-where only.** Without `operators:`, `?price_gte=10` is ignored exactly as before — zero behaviour change for existing apps.
- **Nothing raises at request time**: blank values skipped, unknown bracket keys and non-scalar values ignored (the 1.22 `where_safe?` guard applies to lists). Strict, validated contracts remain Permittable's job.
- **Configuration errors raise at class load**: `operators:` on a `scope:`/`with:` filter, an unknown operator (message lists the valid ones), an unknown `type:`.
- `filterable_rules[field]` now carries `type:`/`operators:` for introspection.
- LIKE escaping uses an explicit `ESCAPE '\'` clause (SQLite has no default escape character) — the Searchable approach.
- `class_methods do` → `module ClassMethods` (RuboCop `Metrics/BlockLength`; Stateable precedent).

## Docs

- `README.md` — two new example lines, an **Operators** table + casting paragraph, notes.
- `docs/concerns/filterable.md` — signature, `type:`/`operators:` config rows, a catalogue example block with the URL → SQL mapping, three Notes bullets.

```
### Added
- **Controllers::Filterable**: `filter_by … operators: true` (or a subset) adds
  comparison operators to direct-where filters — `not`, `gt`, `gte`, `lt`,
  `lte`, `in`, `not_in`, `null`, `contains`, `starts_with` — read as a suffix
  (`?price_gte=10`) or in bracket form (`?price[gte]=10`). Values are cast
  through the column's own type, or through the new `type:` option, which also
  pre-casts the value handed to a `with:` lambda. Opt-in per filter; existing
  filters are unchanged.
```

## Test plan

- [x] +15 examples: plain equality unchanged; gt/gte/lt/lte (suffix and bracket) against a decimal column; not/in/not_in (comma list and array); null true/false/0; contains/starts_with with `%` and `_` escaped (a literal `100%` matches only the row containing it); casting through an integer column; blank/non-scalar/unknown-key handling; no-`operators:` backward compatibility; operator subset; `type:` with `with:` lambdas (Integer and Date); rule introspection; three configuration errors.
- [x] Full suite: 1272 examples, 0 failures (98.35%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/filterable.rb
+++ b/lib/concerns_on_rails/controllers/filterable.rb
@@ -110,7 +110,10 @@ module ConcernsOnRails
 
       def apply_filter(relation, field, value, options)
         if options[:with]
-          options[:with].call(relation, filter_cast(relation, field, value, options))
+          # `type:` pre-casts a with: lambda's value; the column's own type must
+          # NOT, or every existing lambda on a column-backed param silently
+          # starts receiving true / a Time where it used to get "1" / "2020-01-02".
+          options[:with].call(relation, options[:type] ? options[:type].cast(value) : value)
         elsif options[:scope]
           relation.public_send(options[:scope])
         elsif filterable_scalar?(value)
--- a/lib/concerns_on_rails/controllers/filterable.rb
+++ b/lib/concerns_on_rails/controllers/filterable.rb
@@ -1,4 +1,5 @@
 require "active_support/concern"
+require "active_model/type"
 require "concerns_on_rails/support/scalar_param"
 
 module ConcernsOnRails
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/filterable.rb
```diff
@@ -22,34 +39,69 @@ module Controllers
     module Filterable
       extend ActiveSupport::Concern
 
+      LABEL = "ConcernsOnRails::Controllers::Filterable".freeze
+      OPERATORS = %i[not gt gte lt lte in not_in null contains starts_with].freeze
+      COMPARISONS = { gt: :gt, gte: :gteq, lt: :lt, lte: :lteq }.freeze
+      LIKE_ESCAPE = "\\".freeze
```
VSN2015, line 45: Supporting both suffix operators (`price_gte`) and bracket notation (`price[gte]`) accommodates varying frontend query string conventions.

lib/concerns_on_rails/controllers/filterable.rb
```diff
@@ -58,19 +110,106 @@ def filtered(relation)
...
         if options[:with]
-          options[:with].call(relation, value)
+          options[:with].call(relation, filter_cast(relation, field, value, options))
         elsif options[:scope]
           relation.public_send(options[:scope])
         elsif filterable_scalar?(value)
           relation.where(field => value)
+        elsif options[:operators] && value.respond_to?(:each_pair)
+          apply_filter_operator_hash(relation, field, value, options)
         else
```
VSN2015, line 120: Translating operator filters into parameterized Arel queries prevents SQL injection while supporting range and inequality clauses.

VSN2015 (commented):
LGTM! Adding comparison operators (`_gt`, `_gte`, `_lt`, `_lte`, bracket notation) and type coercion transforms Filterable into a versatile query engine without sacrificing SQL injection safety.

PR #48, opened 2026-09-04
Cacheable: etag_with — fold request context (locale, fields, role) into the ETag + automatic Vary

VSN2015:
## Summary

The ETag derived from the resource's cache key alone. When the body also depends on the **request** — `I18n.locale` (Localizable), sparse fieldsets (Includable's `?fields=`), the caller's role — two different representations of one record shared a validator, and a client that switched locale and revalidated got a **304 for a body it had never seen**. `Vary` alone does not fix that: it keys shared caches on the header, but the client's own `If-None-Match` still matches the old ETag.

`etag_with` is the analogue of Rails' class-level `etag { }`:

```ruby
class Api::ArticlesController < ApplicationController
  include ConcernsOnRails::Controllers::Cacheable

  http_cache_actions :index, :show, max_age: 5.minutes, visibility: :public, vary: "Accept"
  etag_with :locale                    # I18n.locale → folded into the ETag, Vary: Accept-Language
  etag_with :requested_fields          # a controller method
  etag_with { current_user&.role }     # a block, instance_exec'd

  def show
    @article = Article.find(params[:id])
    return unless stale_resource?(@article, extras: [params[:include]])   # per-call context too
    render json: @article
  end
end
```

| Source | Value folded in | Default `Vary` |
|---|---|---|
| `:locale` | `I18n.locale` | `Accept-Language` |
| `:format` | `request.format` | `Accept` |
| `:query` | `request.query_string` | — |
| any other `Symbol` | the controller method of that name | — |
| a block | `instance_exec`'d on the controller | — |

**Design points**

- Values are folded into every validator `stale_resource?` / `set_cache_validators` writes: `W/"md5(base-etag | v | v …)"` — still weak, still deterministic. `nil` values are dropped, so an absent context leaves the ETag exactly as before (zero change for apps that don't declare anything).
- Each source contributes its implied `Vary`; `vary:` on the call overrides, `vary: false` suppresses. Written whenever validators are written and **merged, de-duplicated, with the `http_cache_actions` policy** (after_action) — order-independent.
- An explicit `etag:` stays verbatim only when there is no context to fold in (with context it is combined, as Rails does).
- Unsafe requests still get neither validators nor `Vary` (the 1.22 guard).
- A Symbol that is neither a preset nor a controller method raises a clear `ArgumentError` at request time (methods may be defined after the macro); an empty call, a non-Symbol source or a blank `vary:` raise at class load. `cacheable_etag_extras` exposes the declared entries; `cache_etag_extras(extra)` (public) resolves them for the current request so a `cache_etag_for` override can reuse it.

## Docs

- `README.md` — example lines, the resulting `Vary`, an "ETag context" paragraph.
- `docs/concerns/cacheable.md` — new `etag_with` section with the source table, updated method signatures, `cache_etag_extras`, a Notes bullet on why context belongs in the ETag and not only in `Vary`.

```
### Added
- **Controllers::Cacheable**: `etag_with` folds request context into the ETag —
  presets `:locale` / `:format` / `:query`, controller-method Symbols, or a
  block — so locale-, fieldset- or role-dependent representations of one
  resource never share a validator; each source adds its implied `Vary`
  (`Accept-Language`, `Accept`; `vary:` overrides, `vary: false` suppresses),
  merged with the `http_cache_actions` policy. `stale_resource?` /
  `set_cache_validators` gain a per-call `extras:`.
```

## Test plan

- [x] +12 examples: `:locale` changes the ETag and adds `Vary: Accept-Language`; no 304 for a validator minted under another locale (and a 304 under the same one); determinism; block + method sources with no `Vary`; nil ignored; per-call `extras:`; explicit `etag:` combined with context vs verbatim without; `:format`, `vary:` override and `vary: false`; merge with the policy `Vary`, de-duplicated; unsafe request untouched; introspection + three class-load validations; request-time error for an unknown Symbol.
- [x] Full suite: 1269 examples, 0 failures (98.31%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/cacheable.rb
+++ b/lib/concerns_on_rails/controllers/cacheable.rb
@@ -16,6 +16,8 @@ module ConcernsOnRails
     #
     #     http_cache_actions :index, :show, max_age: 5.minutes,
     #                        visibility: :public, vary: "Accept"
+    #     etag_with :locale                       # representation depends on I18n.locale
+    #     etag_with { current_user&.role }        # ...and on who is asking
     #
     #     def show
     #       @article = Article.find(params[:id])
@@ -41,6 +43,12 @@ module ConcernsOnRails
     #   * The 304 is only sent for safe requests (GET/HEAD) and still carries the
     #     validators AND the policy headers (Cache-Control rides the 304, like
     #     Deprecatable's headers ride the 410).
+    #   * `etag_with` (the analogue of Rails' class-level `etag { }`) folds
+    #     request context into the ETag — locale, requested fields, the current
+    #     user's role — so two representations of one resource never share a
+    #     validator, and each source adds its `Vary` header (`:locale` →
+    #     Accept-Language, `:format` → Accept). Per call: `stale_resource?(r,
+    #     extras: [...])`.
     #
     # Notes:
     #   * `no_store: true` overrides everything (emits the lone `no-store`).
@@ -59,8 +67,18 @@ module ConcernsOnRails
       VALID_VISIBILITY = %i[public private].freeze
       SAFE_METHODS = %w[GET HEAD].freeze
 
+      # `etag_with` presets: the value folded into the ETag and the Vary header
+      # it implies. Any other Symbol names a controller method.
+      ETAG_PRESETS = {
+        locale: { value: -> { I18n.locale if defined?(I18n) }, vary: ["Accept-Language"] },
+        format: { value: -> { request.format.to_s if respond_to?(:request) && request.respond_to?(:format) }, vary: ["Accept"] },
+        query: { value: -> { request.query_string if respond_to?(:request) && request.respond_to?(:query_string) }, vary: [] }
+      }.freeze
+
       included do
         class_attribute :cacheable_rules, instance_accessor: false, default: []
+        # [{ source: Symbol | Proc, vary: [header, ...] }] — see etag_with.
+        class_attribute :cacheable_etag_extras, instance_accessor: false, default: []
         # no_store is applied EARLY as well: rescue_from-rendered errors skip
         # after_action entirely, and for a no_store endpoint (tokens, PII) even
         # the error response must not be stored. Positive freshness policies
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/cacheable.rb
```diff
@@ -93,8 +111,43 @@ def http_cache_actions(*actions, visibility: :private, max_age: nil, must_revali
           self.cacheable_rules = cacheable_rules + [rule]
         end
 
+        # Declare request context that shapes the representation and therefore
+        # belongs in the ETag: presets (:locale, :format, :query), Symbols naming
```
VSN2015, line 115: Allowing `etag_with` to evaluate procs or symbols cleanly captures dynamic request-bound context in the generated ETag digest.

lib/concerns_on_rails/controllers/cacheable.rb
```diff
@@ -163,28 +216,42 @@ def apply_http_no_store
...
         http_cache_send_not_modified
         false
       end
 
       # Set the ETag/Last-Modified response headers (no short-circuit). Returns
-      # the computed { etag:, last_modified: } pair.
-      def set_cache_validators(resource = nil, etag: nil, last_modified: nil)
+      # the computed { etag:, last_modified: } pair. Class-level `etag_with`
+      # values and per-call `extras:` are folded into the ETag (an explicit
+      # `etag:` stays verbatim only when there are none), and the sources' Vary
```
VSN2015, line 235: Automatically synchronizing `Vary` headers with evaluated ETag components ensures HTTP intermediate caches do not serve stale representations across varying client contexts.

VSN2015 (commented):
LGTM! Incorporating request context (`etag_with`) into ETags alongside automated `Vary` header management prevents cross-context cache collisions across locales, user roles, and sparse fieldsets.

VSN2015:

```
# PR #48 Cacheable etag_with — VERDICT: MERGE-AFTER-FIX (2 CRITICAL, design-level)
1. [CRITICAL] README.md:1943-1958 + cacheable.rb:17-20 — headline example pairs
   `visibility: :public, max_age: 5.minutes` with `etag_with { current_user&.role }`.
   User-identifying dimension in ETag on a shareable response -> CDN serves the editor's
   body to anonymous callers for 300s. FIX: split the example / use visibility: :private.
2. [CRITICAL] cacheable.rb:129, :297-303 — LIBRARY DEFAULT, not just docs: a non-preset
   source (Symbol/block) defaults to `vary: []`, and `vary: false` zeroes a preset's
   headers, so the ETag varies on a dimension with NO matching Vary, and nothing stops the
   rule being `visibility: :public`. `Vary` cannot express "role"/"current_user" at all, so
   the only correct response for such a source is `private`.
   FIX: when an entry's resolved vary list is empty and source != :query, either raise on a
   :public rule or force http_cache_control_value to emit `private`.
   NOTE: specs LOCK IN the unsafe behaviour — cacheable_spec.rb:259 and :303 assert
   `not_to have_key("Vary")` for exactly these cases; no example covers :public + non-header
   source. Specs must change too.
3. [IMPORTANT] cacheable.rb:243,:297 — automatic Vary is written ONLY from
   set_cache_validators, never from the after_action apply_http_cache_headers (:195), so a
   class-level etag_with silently does not apply to actions that never call
   stale_resource?. FIX: call http_cache_write_etag_vary from apply_http_cache_headers too
   (idempotent — merge_vary appends + uniqs).
4. [IMPORTANT] cacheable.rb:225 — stale_resource? now always passes `extras:` to
   set_cache_validators, a DOCUMENTED PUBLIC method; an app that overrode/wrapped the old
   3-keyword signature raises ArgumentError: unknown keyword :extras at request time.
   FIX: pass extras: only when non-nil, or call it out as breaking.
Non-blocking: the :locale preset always advertises Vary: Accept-Language even when the app
resolves locale from a param/cookie/current_user.
Checked clean: Vary merged not clobbered; unsafe-request guard intact; weak ETag + RFC 7232
precedence untouched; subclass inheritance per house style.
```

PR #49, opened 2026-09-04
Stateable: timestamps: (<state>_at stamping) and per-event before_/after_<event> hooks

VSN2015:
## Summary

Two things every real state machine grows within a week of adoption:

**1. `timestamps:` — "when did it enter this state?"**

```ruby
stateable_by :status, states: %i[draft review published archived], default: :draft,
             timestamps: true,                        # or %i[published archived]
             transitions: { publish: { from: %i[draft review], to: :published }, archive: { to: :archived } }

article.publish!       # status = "published" AND published_at = Time.current — one update!
article.archived!      # direct setters, transition_to! and transition_all stamp too
```

- Stamps `<state>_at = Time.current` in the **same `update!`** as the state change, so there is no window where the state is set but the stamp is not.
- The default state is **not** stamped on create; re-entering a state re-stamps it and leaves the other stamps alone (`published_at` answers "when was it *last* published").
- Columns are checked at class load with a typed `:datetime` migration hint, **after** the states are validated — a typo in `timestamps: %i[publsihed]` reads as an unknown state, not a missing column.
- `Model.stateable_timestamps` lists the stamped states.

**2. Per-event hooks `before_<event>` / `after_<event>`**

```ruby
def before_publish = check_embargo!
def after_publish  = notify_subscribers
```

- Plain instance methods, looked up by the **affixed** event name (`before_status_publish` with `prefix: true`), skipped when undefined.
- Fire inside the generic pair and the same transaction: `before_transition → before_<event> → write → after_<event> → after_transition`. A raising hook rolls the state change back.
- Like the generic hooks, guarded transitions only — direct setters and `transition_to!` bypass them; `transition_all` fires them once per record.

No behaviour change for models that declare neither.

## Docs

- `README.md` — new "Timestamps and per-event hooks" block in the Stateable section.
- `docs/concerns/stateable.md` — `<state>_at` columns row, `timestamps:` and `lock:` config rows (the latter was undocumented in the table), hook order in the `<event>!` row, two Notes bullets.

```
### Added
- **Models::Stateable**: `stateable_by … timestamps: true` (or a list of
  states) stamps `<state>_at = Time.current` in the same write as the state
  change, for guarded events, direct setters, `transition_to!` and
  `transition_all`; the columns are checked at class load. Per-event hooks
  `before_<event>` / `after_<event>` fire inside the generic
  `before_transition` / `after_transition` pair and the same transaction.
```

## Test plan

- [x] +14 examples: stamping via guarded event, direct setter, `transition_to!` and `transition_all`; default not stamped on create; re-entry re-stamps only that state; a state subset leaves unlisted states unstamped; introspection; missing column → typed hint; unknown state and bad option type rejected. Hook order and event isolation; raising `after_<event>` rolls back; affixed hook name (and the un-affixed one is ignored); setters/`transition_to!` bypass; once per record via `transition_all`.
- [x] Full suite: 1271 examples, 0 failures (98.30%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/models/stateable.rb
+++ b/lib/concerns_on_rails/models/stateable.rb
@@ -31,8 +31,20 @@ module ConcernsOnRails
     #
     # Plus a generic `transition_to!(state)`.
     #
-    # Options for stateable_by: default:, transitions:, prefix:, suffix:, lock:
-    # (prefix:/suffix: take `true` to use the field name, or a literal string/symbol).
+    # Per-event hooks: define `before_<event>` / `after_<event>` (affixed like the
+    # event method — `before_status_publish` with prefix: true) and they fire
+    # around that guarded transition, inside the generic before_transition /
+    # after_transition pair and the same transaction.
+    #
+    # `timestamps: true` (or a list of states) stamps `<state>_at = Time.current`
+    # in the same write as the state change — guarded events, direct setters
+    # and transition_to! alike — so "when was it published / archived?" needs no
+    # callback. The columns are checked at macro time (typed :datetime in the
+    # migration hint); the default state is not stamped on create.
+    #
+    # Options for stateable_by: default:, transitions:, prefix:, suffix:, lock:,
+    # timestamps: (prefix:/suffix: take `true` to use the field name, or a
+    # literal string/symbol).
     #
     # Notes:
     #   * String columns only (store the state name) — not integer-backed like Rails enum.
@@ -53,7 +65,7 @@ module ConcernsOnRails
       class InvalidTransition < StandardError; end
 
       # Valid stateable_by keyword options (everything besides field/states:).
-      OPTIONS = %i[default transitions prefix suffix lock].freeze
+      OPTIONS = %i[default transitions prefix suffix lock timestamps].freeze
 
       included do
         class_attribute :stateable_field, instance_accessor: false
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/stateable.rb
```diff
@@ -63,18 +75,21 @@ class InvalidTransition < StandardError; end
         class_attribute :stateable_prefix, instance_accessor: false
```
VSN2015, line 75: Executing before/after event hooks around transitions allows specific event side-effects without bloating state transition declarations.

lib/concerns_on_rails/models/stateable.rb
```diff
@@ -209,29 +247,44 @@ def stateable_apply_default
       # With `lock: true` the guard is re-checked under a row lock (with_lock
       # reloads, so the state read is the committed one) — closing the
       # check-then-write race between two concurrent transitions.
-      def stateable_perform_transition!(field, to, from, event)
+      def stateable_perform_transition!(field, to, from, event, name)
```
VSN2015, line 250: Automatically updating `<state>_at` columns records timestamps for lifecycle tracking and SLA reporting.

VSN2015 (commented):
LGTM! Stamping `<state>_at` timestamps alongside transition callbacks (`before_<event>` / `after_<event>`) delivers complete state transition auditing and hook points.

PR #50, opened 2026-09-04
Auditable: gem-wide audit_actor fallback, actor: false opt-out, Symbol actors

VSN2015:
## Summary

Every audited model had to repeat `actor: -> { Current.user&.email }`, and a model that forgot it silently recorded entries with no `"by"`. Real apps have **one** actor source, so it can now be set once:

```ruby
# config/initializers/concerns_on_rails.rb
ConcernsOnRails.setup do |config|
  config.cache_store = -> { Rails.cache }
  config.audit_actor = -> { Current.user&.id }   # fallback "by" for every Auditable model
end

class Product < ApplicationRecord
  auditable_by :price, :status                  # uses the gem-wide actor
end
class ImportRow < ApplicationRecord
  auditable_by :status, actor: false            # explicitly no "by" for this model
end
class Invoice < ApplicationRecord
  auditable_by :total_cents, actor: :updated_by_id   # a record method / column
end
```

- **`config.audit_actor`** — the fallback for every model that passes no `actor:` (the `cache_store` pattern: class-level wins). Read **per save**, so an initializer that runs after the model class loads still applies; `instance_exec`'d on the record exactly like a model-level actor. Setting a non-callable raises at configuration time.
- **`actor: false`** opts one model out of the fallback.
- **`actor: :symbol`** calls that method on the record — the common "we already store who edited it" case; `"by"` is omitted when it returns `nil`.
- The resolved value is JSON-coerced like any entry value, so `-> { { id: Current.user.id, type: "User" } }` works.
- Validation: an `actor:` that is neither callable, Symbol, `nil` nor `false` raises at class load (the existing "actor must be callable" wording is preserved for anyone matching on it).

No behaviour change for apps that configure nothing.

## Docs

- `README.md` — the `ConcernsOnRails.setup` block gains `audit_actor`; the Auditable example and Options line cover the fallback, `false` and Symbols.
- `docs/concerns/auditable.md` — `actor:` row, the `"by"` note, a new "Gem-wide actor" section, the actor-proc note.

```
### Added
- **Models::Auditable**: `ConcernsOnRails.setup { |c| c.audit_actor = -> {
  Current.user&.id } }` sets the `"by"` actor once for every audited model that
  passes no `actor:` (resolved per save, `instance_exec`'d on the record);
  `actor: false` opts a model out, and `actor: :method_name` stamps a record
  method's value. A non-callable `audit_actor` raises when set.
```

## Test plan

- [x] +9 Auditable examples: fallback used; fallback `instance_exec`'d on the record; model-level `actor:` wins; `actor: false` opts out (and is exposed via `auditable_actor`); Symbol actor with a value and with `nil` (no `"by"`); lazy per-save resolution; Hash coercion; `actor: "nope"` and `audit_actor = "nope"` rejected. +1 Configuration example (round trip, rejection, reset to nil).
- [x] Full suite: 1267 examples, 0 failures (98.35%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/configuration.rb
+++ b/lib/concerns_on_rails/configuration.rb
@@ -14,8 +14,23 @@ module ConcernsOnRails
   # #write(expires_in:, unless_exist:) / #delete for idempotency. There is
   # still no in-process default on purpose — a non-atomic store silently
   # under-counts, so the host must opt in explicitly (just once, here).
+  #
+  # `audit_actor` is the fallback actor for Models::Auditable: a callable,
+  # instance_exec'd on the record at save time, whose value is stamped as
+  # "by" on every audit entry of every model that passes no `actor:` of its
+  # own (`auditable_by ..., actor: -> { ... }` still wins; `actor: false` opts
+  # a model out). Typically `-> { Current.user&.id }`.
   class Configuration
     attr_accessor :cache_store
+    attr_reader :audit_actor
+
+    def audit_actor=(value)
+      unless value.nil? || value.respond_to?(:call)
+        raise ArgumentError, "ConcernsOnRails.config.audit_actor must be callable (respond to #call) or nil"
+      end
+
+      @audit_actor = value
+    end
 
     # The fallback store with any callable resolved (per lookup, so a Proc
     # reading Rails.cache follows a swapped-out cache in tests). nil when the
--- a/lib/concerns_on_rails/models/auditable.rb
+++ b/lib/concerns_on_rails/models/auditable.rb
@@ -27,6 +27,12 @@ module ConcernsOnRails
     #   product.audited_changes_since(1.day.ago)
     #   product.clear_audit_trail!                 # wipe the column (skips callbacks)
     #
+    # Actor resolution ("by"): a model's `actor:` (a callable instance_exec'd
+    # on the record, or a Symbol naming a record method) wins; otherwise the
+    # gem-wide fallback `ConcernsOnRails.setup { |c| c.audit_actor = -> {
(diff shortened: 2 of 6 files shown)


lib/concerns_on_rails/configuration.rb
```diff
@@ -14,8 +14,23 @@ module ConcernsOnRails
   # #write(expires_in:, unless_exist:) / #delete for idempotency. There is
   # still no in-process default on purpose — a non-atomic store silently
   # under-counts, so the host must opt in explicitly (just once, here).
+  #
+  # `audit_actor` is the fallback actor for Models::Auditable: a callable,
+  # instance_exec'd on the record at save time, whose value is stamped as
+  # "by" on every audit entry of every model that passes no `actor:` of its
```
VSN2015, line 20: Adding global `audit_actor` configuration provides a single point to declare current user resolution (e.g. `Current.user`).

lib/concerns_on_rails/models/auditable.rb
```diff
@@ -103,7 +109,14 @@ def validate_auditable!(fields, into:, actor:, max_entries:, max_value_length:)
           unless positive_integer_or_nil?(max_value_length)
             raise ArgumentError, "#{LABEL}: max_value_length must be a positive Integer or nil"
```
VSN2015, line 110: Allowing models to opt out with `actor: false` or override with symbol methods provides clean model-level customization.

VSN2015 (commented):
LGTM! Gem-wide `audit_actor` configuration reduces repetitive per-model actor definitions, while `actor: false` and symbol method support offer clean escape hatches.

PR #51, opened 2026-09-04
Tokenizable: expires_in: (expiring tokens) and consume_<field> (race-safe single use)

VSN2015:
## Summary

Password resets, magic links and invite codes are tokens that must **stop working after a while** and **work exactly once**. Both were left to the host app. Now:

```ruby
tokenizable_by :reset_password_token, length: 24, expires_in: 2.hours   # needs reset_password_token_expires_at (datetime)
tokenizable_by :invite_code, type: :alphanumeric, length: 8

User.authenticate_by_reset_password_token(token)   # nil once expired
User.consume_reset_password_token(token)           # authenticate AND revoke atomically; nil the second time
User.consume_invite_code(code)                     # single use works for any field
User.reset_password_token_expired                  # scope for cleanup jobs
user.reset_password_token_expired?
```

**`expires_in:`**
- Stamps `<field>_expires_at = Time.current + expires_in` on create, on `regenerate_<field>!`, and for a caller-supplied token that arrives without its own expiry (a caller-set expiry is respected); `revoke_<field>!` clears it. The column is checked at class load with a typed migration hint.
- `authenticate_by_<field>` returns `nil` for an expired token — the security half. `<field>_expired?` reports it; a `nil` expiry (rows from before the option existed) never expires; a field without `expires_in:` is never expired.
- `<field>_expired` scope: rows whose expiry has passed.

**`consume_<field>(value)`** — defined for every field:
- Authenticates (constant-time compare, expiry-aware), then revokes token + expiry with a **conditional `UPDATE … WHERE id = ? AND <field> = ?`**. Returns the reloaded record only when exactly one row was revoked, so two concurrent consumers cannot both win — the loser gets `nil`. An expired or wrong token is refused and left untouched.

Internals: the three `class_methods do` blocks became one `module ClassMethods` (Stateable precedent) and the label became a constant. No behaviour change for existing fields beyond gaining `consume_<field>` and `<field>_expired?` (always `false` without `expires_in:`).

## Docs

- `README.md` — example block, `expires_in:` option row, three Notes bullets.
- `docs/concerns/tokenizable.md` — columns row, `expires_in:` config row, Scopes section, `<field>_expired?` / `consume_<field>` / `tokenizable_expiry_column` rows, two Notes bullets.

```
### Added
- **Models::Tokenizable**: `tokenizable_by … expires_in:` gives a token a
  lifetime — `<field>_expires_at` is stamped on every generation,
  `authenticate_by_<field>` refuses an expired token, `<field>_expired?` and
  the `<field>_expired` scope report it. New `consume_<field>(value)` (every
  field) authenticates and revokes in one race-safe step for single-use
  tokens (password resets, magic links, invite codes).
```

## Test plan

- [x] +13 examples: stamping on create and regenerate; caller-supplied token gets the lifetime, caller-set expiry respected; revoke clears both; `expired?` at the boundary, with a `nil` expiry and on a non-expiring field; `authenticate_by_` refuses expired; consume happy path and second call `nil`; the race (a stubbed interleaving revokes between authenticate and UPDATE → `nil`); wrong/blank/expired refused and left in place; consume on a non-expiring field; the scope (and its absence on non-expiring fields); config introspection; column requirement with typed hint; non-positive `expires_in:` rejected.
- [x] Full suite: 1270 examples, 0 failures (98.34%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/models/tokenizable.rb
+++ b/lib/concerns_on_rails/models/tokenizable.rb
@@ -13,7 +13,7 @@ module ConcernsOnRails
     #     include ConcernsOnRails::Tokenizable
     #
     #     tokenizable_by :api_token                              # 32-char URL-safe
-    #     tokenizable_by :reset_password_token, length: 24
+    #     tokenizable_by :reset_password_token, length: 24, expires_in: 2.hours
     #     tokenizable_by :invite_code, type: :alphanumeric, length: 8
     #   end
     #
@@ -24,6 +24,16 @@ module ConcernsOnRails
     #
     #   User.find_by_api_token(token)             # Rails default
     #   User.authenticate_by_api_token(token)     # constant-time compare; returns record or nil
+    #   User.consume_invite_code(code)            # authenticate AND revoke in one step (single use)
+    #
+    # `expires_in:` gives a field a lifetime: `<field>_expires_at` (a datetime
+    # column you add) is stamped whenever the token is generated — on create,
+    # on regenerate_<field>!, and for a caller-supplied token that arrives
+    # without its own expiry — and cleared by revoke_<field>!.
+    # `authenticate_by_<field>` / `consume_<field>` refuse an expired token,
+    # `<field>_expired?` reports it, and the `<field>_expired` scope finds rows
+    # for cleanup. `consume_<field>` (every field) revokes with a conditional
+    # UPDATE, so two concurrent consumers cannot both win.
     #
     # Unlike Hashable, one model can declare multiple token fields, generation is
     # URL-safe by default, and `assign_tokenizable_value` retries on uniqueness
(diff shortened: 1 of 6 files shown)


lib/concerns_on_rails/models/tokenizable.rb
```diff
@@ -34,6 +44,7 @@ module Models
     module Tokenizable
       extend ActiveSupport::Concern
```
VSN2015, line 45: Automatic expiration tracking via `expires_in:` handles time-bounded token validation out of the box.

lib/concerns_on_rails/models/tokenizable.rb
```diff
@@ -150,9 +186,46 @@ def tokenizable_unique_value(field)
...
         end
 
-        raise "ConcernsOnRails::Models::Tokenizable: could not generate a unique value for '#{field}' " \
+        raise "#{LABEL}: could not generate a unique value for '#{field}' " \
               "after #{MAX_GENERATION_ATTEMPTS} attempts — consider a longer length or a larger alphabet"
       end
+
+      private
+
+      # A fresh token plus, for an expiring field, a fresh expiry.
```
VSN2015, line 195: Implementing `consume_<field>` with atomic SQL clearing prevents race conditions during token redemption.

VSN2015 (commented):
LGTM! Expiring tokens (`expires_in:`) and atomic single-use consumption (`consume_<field>`) cover critical security primitives for reset tokens and magic links.

PR #52, opened 2026-09-04
SoftDeletable: cascade: soft-deletes and restores has_many/has_one dependents

VSN2015:
## Summary

Soft-deleting a post left its comments and cover live — and `dependent: :destroy` hard-deletes, the wrong tool for a soft-delete model. Now:

```ruby
class Post < ApplicationRecord
  include ConcernsOnRails::SoftDeletable
  has_many :comments      # Comment includes SoftDeletable (and may cascade further, e.g. to likes)
  has_one  :cover
  soft_deletable_by :deleted_at, cascade: %i[comments cover]
end

post.soft_delete!              # comments + cover soft-deleted in the same transaction, with the post's exact timestamp
post.restore!                  # brings back what the cascade deleted — not a comment someone trashed last week
post.soft_delete!(at: 1.day.ago)   # new at: keyword: backdate, or hand a timestamp down a cascade
```

**Design**

- **Through the dependents' own `soft_delete!` / `restore!`** — their hooks and their own `cascade:` run (recursive), and a raising dependent rolls the parent back (one transaction).
- **Exact-timestamp matching for restore.** The cascade hands the parent's timestamp down via `soft_delete!(at:)`; `restore!` restores only dependents whose `deleted_at` equals the parent's. A dependent deleted independently earlier keeps its own timestamp and stays deleted — no extra columns, no time-window heuristic. (Give the columns `precision: 6`, the Rails 7 default, if two parents may be deleted within one second.)
- **Validation.** `has_many` / `has_one` only — `belongs_to`, HABTM and `:through` raise `ArgumentError` at class load. The target model must include SoftDeletable: checked at class load when the class already resolves, otherwise on the first cascade (Rails' `compute_type` cannot resolve association targets from inside an anonymous or not-yet-loaded class body).
- **Batch verbs cascade too:** with a `cascade:` configured the single-`UPDATE` fast paths of `soft_delete_all` / `restore_all` are disabled (a bulk `UPDATE` cannot follow associations), so they stream per record.
- `soft_delete_cascade` exposes the configuration. No behaviour change for models without `cascade:`.

## Docs

- `README.md` — a "Cascading to dependents" block in the SoftDeletable section.
- `docs/concerns/soft-deletable.md` — `cascade:` config row, full macro signature, `soft_delete!(at:)` / `restore!` rows, two Notes bullets (timestamp matching + precision/`travel_to`, lazy target resolution).

```
### Added
- **Models::SoftDeletable**: `soft_deletable_by … cascade: %i[comments cover]`
  soft-deletes has_many / has_one dependents with the record (same
  transaction, same timestamp, through their own `soft_delete!` so hooks and
  nested cascades run) and restores exactly those on `restore!` — a dependent
  deleted independently earlier stays deleted. New `soft_delete!(at:)` keyword.
  With a cascade configured, `soft_delete_all` / `restore_all` take the
  per-record path.
```

## Test plan

- [x] +9 examples: recursive cascade (post → comments → likes, plus has_one cover) with an exact sub-second timestamp; dependents' hooks; `restore!` restores cascaded dependents but not one deleted independently earlier; pre-deleted dependents keep their timestamp; `at:` backdating standalone and through the cascade; a raising dependent rolls the parent back; `soft_delete_all` / `restore_all` cascade; introspection; unknown association / `belongs_to` / `:through` / non-SoftDeletable target rejected at class load (named class).
- [x] Full suite: 1266 examples, 0 failures (98.34%).
- [x] RuboCop clean.


lib/concerns_on_rails/models/soft_deletable.rb
```diff
@@ -45,11 +48,23 @@ module ClassMethods
         # Example:
         #   soft_deletable_by :deleted_at, touch: false
         #   soft_deletable_by :deleted_at, default_scope: false  # don't hide deleted rows from .all
-        def soft_deletable_by(field = nil, touch: true, default_scope: true, prefix: nil, suffix: nil)
+        #   soft_deletable_by :deleted_at, cascade: %i[comments attachments]
+        #
+        # `cascade:` names has_many / has_one associations whose records are
+        # soft-deleted with the parent (inside its transaction, with the
+        # parent's exact timestamp, through their own soft_delete! — hooks and
```
VSN2015, line 55: Supporting `cascade:` configuration allows declarative cascading of soft-delete and restore operations down association trees.

lib/concerns_on_rails/models/soft_deletable.rb
```diff
@@ -187,6 +202,47 @@ def soft_delete_batch_fast_path?(kind)
                     end
           ConcernsOnRails::Support::BatchOps.unoverridden?(self, ConcernsOnRails::Models::SoftDeletable, *methods)
         end
+
```
VSN2015, line 205: Wrapping cascading soft-deletes and restores in a transaction ensures atomic state updates across parent and child models.

VSN2015 (commented):
LGTM! The `cascade:` option solves the longstanding gap where soft-deleting parents left dependent records orphaned without resorting to destructive `dependent: :destroy`.

PR #53, opened 2026-09-04
Localizable: Content-Language response header + Vary: Accept-Language

VSN2015:
## Summary

A localized JSON body is a different representation per locale, yet the response said nothing about it — no `Content-Language` for clients, and no `Vary` for shared caches, so a CDN or `Rack::Cache` in front of the API could serve one client's French to the next client's English.

`switch_locale` now writes, **before** running the action (so a `rescue_from`-rendered error carries them too):

```
Content-Language: fr            # the resolved locale, BCP 47 form (pt_BR → pt-BR)
Vary: Accept, Accept-Language   # Accept-Language appended when the header is a locale source
```

- `Vary` is appended to any existing value and de-duplicated, never clobbered (the same rule as Cacheable and the paginators' `Link` header). It is skipped with `header: false`, when the locale can only come from the URL param and the URL already differs.
- `localizable … response_headers: false` disables both headers.
- Guarded on a real response object; no behaviour change to locale resolution itself.

## Docs

- `README.md` — a paragraph on the two headers + the `response_headers:` option.
- `docs/concerns/localizable.md` — config row, `switch_locale` row, macro signature, a Notes bullet on why the headers matter for shared caches.

```
### Added
- **Controllers::Localizable**: every response now carries
  `Content-Language: <resolved locale>` and, when `Accept-Language` is a locale
  source, `Vary: Accept-Language` (appended to an existing `Vary`), written
  before the action so rescued errors carry them too. Opt out with
  `localizable … response_headers: false`.
```

## Test plan

- [x] +7 examples: `Content-Language` set; `Vary` merged with an existing `Accept` and de-duplicated; `header: false` → no `Vary`; `response_headers: false` → neither; `pt_BR` → `pt-BR`; headers survive a raising action; no-op on an object without a response.
- [x] Full suite: 1264 examples, 0 failures (98.33%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/localizable.rb
+++ b/lib/concerns_on_rails/controllers/localizable.rb
@@ -18,9 +18,15 @@ module ConcernsOnRails
     # against `I18n.available_locales` before use, so a stray param or a
     # mismatched `available:` list can never raise `I18n::InvalidLocale`.
     #
+    # Every response carries `Content-Language: <resolved locale>` (BCP 47
+    # form, `pt_BR` → `pt-BR`) and, when the header is a locale source,
+    # `Vary: Accept-Language` appended to any existing Vary — written before
+    # the action runs, so a rescued error still carries them. Both are behind
+    # `response_headers:` (default `true`).
+    #
     # Options: `available:` (allow-list for param/header matching; defaults to
     # `I18n.available_locales`), `default:`, `param:` (default `:locale`),
-    # `header:` (default `true`).
+    # `header:` (default `true`), `response_headers:` (default `true`).
     module Localizable
       extend ActiveSupport::Concern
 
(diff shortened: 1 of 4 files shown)


VSN2015 (commented):
LGTM! Setting `Content-Language` and `Vary: Accept-Language` headers is critical for HTTP cache correctness (CDN / reverse proxies) and client localization transparency.

lib/concerns_on_rails/controllers/localizable.rb
```diff
@@ -18,9 +18,15 @@ module Controllers
     # against `I18n.available_locales` before use, so a stray param or a
     # mismatched `available:` list can never raise `I18n::InvalidLocale`.
     #
+    # Every response carries `Content-Language: <resolved locale>` (BCP 47
+    # form, `pt_BR` → `pt-BR`) and, when the header is a locale source,
+    # `Vary: Accept-Language` appended to any existing Vary — written before
+    # the action runs, so a rescued error still carries them. Both are behind
+    # `response_headers:` (default `true`).
```
VSN2015, line 25: Emitting `Content-Language` directly indicates the negotiated response language to API consumers and browsers.

lib/concerns_on_rails/controllers/localizable.rb
```diff
@@ -60,6 +70,29 @@ def resolved_locale
 
       private
 
+      # Content-Language always; Vary: Accept-Language only when the header can
+      # influence the choice (a param-only setup already differs by URL).
+      # Vary is appended and de-duplicated, never clobbered (Cacheable, the
```
VSN2015, line 75: Adding `Vary: Accept-Language` guarantees downstream HTTP caches will not serve cached localized content to users with different locale preferences.

PR #54, opened 2026-09-04
Expirable: before/after_expire lifecycle hooks, expire_in!, clear_expiry!

VSN2015:
## Summary

Expirable was the one lifecycle concern **without hooks**: when a trial or a share link expired (`expire!` / `expire_all`) there was no seam to downgrade the account or notify anyone. Following the SoftDeletable / Publishable pattern:

```ruby
class Trial < ApplicationRecord
  include ConcernsOnRails::Expirable
  expirable_by

  def after_expire = account.downgrade!     # fires for expire!, expire_in!, expire_all
end

trial.expire_in!(14.days)    # absolute lifetime from now, whatever the current expiry
trial.clear_expiry!          # never expires
```

- **`before_expire` / `after_expire`** — fired by `expire!` (and so by `expire_in!` and `expire_all`) inside one transaction with the write, so a raising `after_expire` rolls the expiry back; a write that fails validation returns `false` and skips `after_expire`. Not fired by `extend_expiry!` (a renewal) or `clear_expiry!`.
- **Batch fast path stays honest**: the `expire_all` gate now checks the hooks too — overridden, it streams per record so the hooks run for every row; unoverridden, it stays a single `UPDATE` (spec asserts one statement).
- **`expire_in!(duration)`** — `expire!(now + duration)`: an absolute lifetime, where `extend_expiry!(by:)` adds to a future expiry.
- **`clear_expiry!`** — nil expiry, no hooks.

No behaviour change for models that override nothing.

## Docs

- `README.md` — mutators block (`expire_in!`, `clear_expiry!`) and a "Lifecycle hooks" paragraph.
- `docs/concerns/expirable.md` — `expire!` row updated, three new method rows, a Notes bullet.

```
### Added
- **Models::Expirable**: `before_expire` / `after_expire` lifecycle hooks (fired
  by `expire!`, `expire_in!` and `expire_all`, in one transaction with the
  write; overriding them moves `expire_all` to the per-record path), plus
  `expire_in!(duration)` for an absolute lifetime from now and `clear_expiry!`.
```

## Test plan

- [x] +7 examples: hooks fire around `expire!` and `expire_in!` but not `extend_expiry!`/`clear_expiry!`; a raising `after_expire` rolls back; a failed validation returns `false` and skips `after_expire`; `expire_all` per record with an overridden hook (once per row) and a single `UPDATE` without; `expire_in!` from a future and a past expiry; `clear_expiry!`.
- [x] Full suite: 1264 examples, 0 failures (98.33%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/models/expirable.rb
+++ b/lib/concerns_on_rails/models/expirable.rb
@@ -31,9 +31,10 @@ module ConcernsOnRails
         end
 
         # Expire every currently-active record in the relation. Returns the
-        # Integer count. Expirable defines no lifecycle hooks, so this is a
-        # single UPDATE unless the model overrode `expire!` or declares
-        # validations (see Support::BatchOps.fast_path?).
+        # Integer count. A single UPDATE unless the model overrode `expire!`
+        # or a lifecycle hook (before_expire / after_expire), or declares
+        # validations (see Support::BatchOps.fast_path?) — then it streams
+        # per record so the hooks run.
         def expire_all(time = Time.zone.now)
           active = all.public_send(expirable_scope_names.fetch(:active))
           if expirable_batch_fast_path?
@@ -52,11 +53,11 @@ module ConcernsOnRails
         private
 
         # Whether the single-UPDATE fast path is safe — the whole decision
-        # (bang method unoverridden AND the model declares no validations,
-        # plus why) lives in Support::BatchOps.fast_path?. Expirable defines
-        # no lifecycle hooks, so `expire!` is the only method to check.
+        # (bang method and hooks unoverridden AND the model declares no
+        # validations, plus why) lives in Support::BatchOps.fast_path?.
         def expirable_batch_fast_path?
-          ConcernsOnRails::Support::BatchOps.fast_path?(self, ConcernsOnRails::Models::Expirable, :expire!)
+          ConcernsOnRails::Support::BatchOps.fast_path?(self, ConcernsOnRails::Models::Expirable,
+                                                        :expire!, :before_expire, :after_expire)
         end
 
         # Scopes live here (not in `included do`) so their names can be affixed —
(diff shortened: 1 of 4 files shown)


VSN2015 (commented):
LGTM! Introducing `before_expire` / `after_expire` lifecycle hooks alongside `expire_in!` and `clear_expiry!` completes Expirable's lifecycle management features.

lib/concerns_on_rails/models/expirable.rb
```diff
@@ -31,9 +31,10 @@ def expirable_by(field = DEFAULT_FIELD, prefix: nil, suffix: nil)
         end
 
         # Expire every currently-active record in the relation. Returns the
-        # Integer count. Expirable defines no lifecycle hooks, so this is a
-        # single UPDATE unless the model overrode `expire!` or declares
-        # validations (see Support::BatchOps.fast_path?).
+        # Integer count. A single UPDATE unless the model overrode `expire!`
+        # or a lifecycle hook (before_expire / after_expire), or declares
```
VSN2015, line 35: Registering ActiveModel callbacks for `:expire` allows applications to tie side effects (such as notifying users or downgrading privileges) to expiration events.

lib/concerns_on_rails/models/expirable.rb
```diff
@@ -96,8 +97,36 @@ def expired?
...
+      # `expire_in!` and `expire_all`), not by `extend_expiry!` (a renewal) or
+      # `clear_expiry!`. Overriding either one moves `expire_all` to the
+      # per-record path.
+      def before_expire; end
+      def after_expire; end
+
+      # Write the expiry (default: now, i.e. expire immediately). The hooks and
+      # the write share one transaction, so a raising after_expire rolls the
+      # expiry back (SoftDeletable's pattern); a failed write (validation)
+      # returns false and skips after_expire.
```
VSN2015, line 110: The `expire_in!` helper method provides clean ergonomics for extending or setting relative expiry windows.

PR #58, opened 2026-09-04
Schedulable: overlapping(from, to) scope and overlaps? predicate

VSN2015:
## Summary

"Which bookings clash with this one?" and "what's on the calendar between Monday and Friday?" are the two Schedulable queries every app writes by hand — and usually gets the boundaries wrong on.

```ruby
Room.overlapping(from, to)        # windows intersecting [from, to)
Room.overlapping(from..to)        # Range form; `..` makes the end inclusive, `...` keeps it exclusive
Room.overlapping(from, nil)       # nil on either side = unbounded
booking.overlaps?(from, to)       # instance-side mirror, same rules
```

- **The standard interval test** — `starts_at < to AND (ends_at IS NULL OR ends_at > from)` — with the same inclusive-start / exclusive-end rules as `active_at`, so back-to-back bookings (09–10, 10–11) do not clash and a window that merely touches the query window is excluded.
- Unstarted records (`nil` start) never overlap, matching `active_at`. Works with an `ends_at`-only configuration.
- Affixable like the other scopes (added to `SCOPE_BASES`), chainable, and an inverted window raises `ArgumentError`.
- The window normalization and the relation builder are public class methods (`schedulable_window`, `schedulable_overlapping`) because scope lambdas resolve methods through the relation, which cannot reach private class methods.

## Docs

- `README.md` — three lines in the Schedulable example.
- `docs/concerns/schedulable.md` — `.overlapping` scope row, `overlaps?` predicate row, a Notes bullet on the interval test.

```
### Added
- **Models::Schedulable**: `.overlapping(from, to)` (or a Range; `nil` sides
  unbounded) returns records whose window intersects `[from, to)` with the
  same inclusive-start / exclusive-end rules as `active_at`, and
  `#overlaps?(from, to)` mirrors it on a record. Affixable like the other
  scopes.
```

## Test plan

- [x] +9 examples: intersection set; touching boundaries excluded; Range with inclusive vs exclusive end; `nil` sides (both, and each); unstarted records excluded; chainable + inverted window raises; predicate mirror across all cases; affixing (`promo_overlapping`); `ends_at`-only model.
- [x] Full suite: 1266 examples, 0 failures (98.34%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/models/schedulable.rb
+++ b/lib/concerns_on_rails/models/schedulable.rb
@@ -9,7 +9,7 @@ module ConcernsOnRails
 
       DEFAULT_STARTS_AT_FIELD = :starts_at
       DEFAULT_ENDS_AT_FIELD = :ends_at
-      SCOPE_BASES = %i[active_at current upcoming expired].freeze
+      SCOPE_BASES = %i[active_at current upcoming expired overlapping].freeze
 
       included do
         class_attribute :schedulable_starts_at_field, instance_accessor: false, default: DEFAULT_STARTS_AT_FIELD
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/schedulable.rb
```diff
@@ -49,8 +49,50 @@ def schedulable_by(starts_at: DEFAULT_STARTS_AT_FIELD, ends_at: DEFAULT_ENDS_AT_
...
 
+        # The relation behind the `overlapping` scope. Public (like
+        # schedulable_window) because scope lambdas resolve methods through the
+        # relation, which cannot reach private class methods.
+        def schedulable_overlapping(from, to = nil)
+          from, to, inclusive_end = schedulable_window(from, to)
+          relation = all
+          relation = schedulable_started_before(relation, to, inclusive_end) if schedulable_starts_at_field
+          relation = schedulable_ending_after(relation, from) if schedulable_ends_at_field && from
+          relation
```
VSN2015, line 60: The `overlapping` SQL scope handles both open-ended boundaries (NULL starts/ends) and closed intervals correctly with proper boundary conditions.

lib/concerns_on_rails/models/schedulable.rb
```diff
@@ -100,6 +150,13 @@ def current?
...
       end
 
+      # Does this record's window intersect [from, to)? Mirrors the
+      # `overlapping` scope — boundaries, nil sides and Ranges included.
+      def overlaps?(from, to = nil)
+        from, to, inclusive_end = self.class.schedulable_window(from, to)
+        schedulable_starts_before?(to, inclusive_end) && schedulable_ends_after?(from)
+      end
+
       def upcoming?
```
VSN2015, line 160: The `overlaps?` predicate accepting either explicit `(from, to)` timestamps or another schedulable record makes booking conflict validation intuitive.

VSN2015 (commented):
LGTM! Adding `overlapping(from, to)` scope and `overlaps?` predicate resolves the most error-prone calendar and booking conflict calculations cleanly.

PR #59, opened 2026-09-04
CursorPaginatable: signed: — tamper-proof cursors (HMAC-SHA256, fail closed)

VSN2015:
## Summary

The README has documented since the concern shipped that cursors are **not signed**: a client can decode one and mint another position. Harmless for public feeds (values are cast through the model's types and bound by Arel; the relation's scoping still applies) — not for feeds where the position itself is sensitive.

```ruby
cursor_paginate_by order: { created_at: :desc }, signed: true            # key = Rails.application.secret_key_base
cursor_paginate_by order: { created_at: :desc }, signed: "a-long-key"    # explicit key
cursor_paginate_by order: { created_at: :desc }, signed: -> { Current.tenant.cursor_key }   # per request, rotatable
```

- **Token shape**: `<url-safe base64 payload>.<hex HMAC-SHA256>` — the dot never appears in the payload alphabet, so the split is unambiguous and the token stays URL-safe. Prev cursors are signed too.
- **Fail closed**: `?cursor=` is verified in constant time (`SecurityUtils.secure_compare`) **before** the payload is parsed; a tampered payload, a wrong signature, a trailing segment, or an unsigned token on a signed endpoint is an `InvalidCursor` → the existing auto-rescued 400. No mixed mode by design, so turning signing on invalidates in-flight cursors (clients restart from page one) — documented.
- `signed: true` outside a Rails app raises `ArgumentError` with a hint to pass a callable; a blank key or another type is rejected at class load.
- Unsigned remains the default and existing tokens keep their exact shape. Configuration is untouched (kept out of this PR to avoid a conflict with #50).

## Docs

- `README.md` — the "not signed" note now describes `signed:`.
- `docs/concerns/cursor-paginatable.md` — `signed` config row, a Notes bullet.

```
### Added
- **Controllers::CursorPaginatable**: `cursor_paginate_by … signed: true` (or a
  String key / callable) appends a URL-safe HMAC-SHA256 to every cursor and
  rejects tampered, mis-signed or unsigned tokens with the usual 400, so
  clients can no longer hand-craft page positions. Unsigned stays the default.
```

## Test plan

- [x] +8 examples: token shape and an identical page walk; five rejection shapes (forged payload, flipped signature, unsigned, empty signature, trailing segment); signed prev cursors in bidirectional mode; Proc and String keys with cross-key rejection; `secret_key_base` via a stubbed `Rails`; missing Rails → configuration error; class-load validation of `signed:`; unsigned default and shape.
- [x] Full suite: 1265 examples, 0 failures (98.33%).
- [x] RuboCop clean.


lib/concerns_on_rails/controllers/cursor_paginatable.rb
```diff
@@ -174,9 +179,19 @@ def cursor_paginate_by(order: nil, order_presets: nil, default_preset: nil, orde
           self.cursor_paginatable_bidirectional = bidirectional ? true : false
           self.cursor_paginatable_predicate = CursorPaginatable.validate_predicate!(predicate)
           self.cursor_paginatable_link_header = link_header ? true : false
+          self.cursor_paginatable_signed = CursorPaginatable.validate_signed!(signed)
         end
       end
```
VSN2015, line 185: Adding opt-in `signed: true` enables HMAC-SHA256 signature verification while maintaining backwards compatibility for existing deployments.

lib/concerns_on_rails/controllers/cursor_paginatable.rb
```diff
@@ -339,7 +354,55 @@ def encode_cursor(model, pairs, record, direction)
           "d" => direction,
           "v" => pairs.map { |col, _dir| serialize_cursor_value(cursor_boundary_value!(record, col)) }
         }
-        cursor_base64_encode(JSON.generate(payload))
+        token = cursor_base64_encode(JSON.generate(payload))
+        cursor_signing_key ? "#{token}.#{cursor_signature(token)}" : token
+      end
+
```
VSN2015, line 360: Failing closed on tampered or corrupted cursor signatures prevents invalid cursor parameters from generating unhandled 500 errors.

VSN2015 (commented):
LGTM! HMAC-SHA256 signed cursors ensure tamper-proof cursor pagination, preventing clients from fabricating pagination state or bypassing access boundaries.

PR #60, opened 2026-09-04
WebhookVerifiable: replay protection (replay: / replay_ttl:) for timestamp-less schemes

VSN2015:
## Summary

GitHub, Shopify and plain hex/base64 signatures carry no timestamp, so a **captured delivery verifies forever** — Stripe's `tolerance:` was the only replay defence, and only for Stripe.

```ruby
verify_webhook :github, secret: ENV["GH_SECRET"], scheme: :github, replay: true            # gem-wide cache_store
verify_webhook :shop,   secret: ENV["SHOP_SECRET"], scheme: :shopify, replay: Rails.cache, replay_ttl: 6.hours
```

- **After** a signature verifies, `SHA256(signature header)` is written to the store with `unless_exist:` — atomic (memcached `add` / Redis `SET NX` through `Rails.cache`), so two concurrent replays lose exactly one. A second delivery with the same signature within `replay_ttl:` (default 24 hours) is rejected with **409 `webhook_replayed`** through the existing failure funnel (`webhook_verification_failed`, Respondable-aware, overridable).
- Runs **only after verification**, so forged traffic never consumes a slot; keys are scoped per `controller#action`.
- `replay: true` uses `ConcernsOnRails.config.cache_store` (the Throttleable / Idempotentable pattern, same setup hint when unset); any object responding to `#write(key, value, expires_in:, unless_exist:)` works.
- Semantics: two legitimate deliveries with different bodies pass; Stripe retries re-sign with a fresh `t=` and pass; a byte-identical redelivery (a replayed capture, or a manual GitHub "Redeliver") is the rejected case — documented with the override for teams whose handlers are idempotent and prefer answering 200.
- Validation at class load: `replay_ttl:` needs `replay:`; `replay:` must be `true` or a store; the ttl must be positive. Rules expose `replay` / `replay_ttl`.

## Docs

- `README.md` — options line, a "Replay protection" note, the failure-code list.
- `docs/concerns/webhook-verifiable.md` — `replay:` / `replay_ttl:` rows, a Notes bullet.

```
### Added
- **Controllers::WebhookVerifiable**: `verify_webhook … replay: true` (or a
  store) + `replay_ttl:` rejects a byte-identical redelivery with 409
  `webhook_replayed` — replay protection for the schemes that carry no
  timestamp (GitHub, Shopify, plain HMAC). The check runs only after the
  signature verifies and uses an atomic `unless_exist` write.
```

## Test plan

- [x] +9 examples: first delivery accepted, identical second rejected (409, `webhook_verified?` false); different body passes; forged delivery writes nothing; key is a digest (raw signature never stored) with `unless_exist` and the default ttl, custom ttl honoured; per-action scoping; `replay: true` raises the setup hint without a store and works with the gem-wide one; Stripe header replay rejected; three class-load validations; rule introspection.
- [x] Full suite: 1266 examples, 0 failures (98.33%).
- [x] RuboCop clean.


lib/concerns_on_rails/controllers/webhook_verifiable.rb
```diff
@@ -93,16 +94,28 @@ module ClassMethods
         # Declare signature verification for the given actions (none =
         # catch-all). Each call appends a rule; the FIRST rule matching the
         # current action wins, so declare specific rules before a catch-all.
-        def verify_webhook(*actions, secret:, scheme: :hex, header: nil, tolerance: nil, digest: :sha256)
+        #
+        # `replay:` adds replay protection for schemes that carry no timestamp
+        # (GitHub, Shopify, plain hex/base64 — Stripe's `tolerance:` only bounds
+        # the window): after a signature verifies, a digest of the signature
```
VSN2015, line 100: Allowing configurable replay TTLs provides an effective defense against webhook replay attacks on schemes without native timestamps.

lib/concerns_on_rails/controllers/webhook_verifiable.rb
```diff
@@ -210,6 +244,37 @@ def webhook_render_outcome(rule, outcome)
         end
       end
 
+      # Replay check, run ONLY after the signature verified so forged traffic
+      # never consumes a slot. The key is a SHA256 of the signature header
+      # (for a valid signature that identifies the body, and for Stripe the
+      # attempt's timestamp too), scoped per controller#action; the atomic
```
VSN2015, line 250: Tracking payload signatures atomically in Rails cache prevents race conditions when duplicate webhook requests arrive concurrently.

VSN2015 (commented):
LGTM! Adding cache-backed replay protection (`replay:` / `replay_ttl:`) closes a critical security vulnerability for webhook providers like GitHub and Shopify that lack timestamp headers in their payloads.

PR #61, opened 2026-09-04
Lockable: unlock_token: — self-service unlock links (Devise :email strategy, minus the mailer)

VSN2015:
## Summary

"Your account is locked — click here to unlock it" was the one Lockable flow that still needed Devise. Now:

```ruby
class User < ApplicationRecord
  include ConcernsOnRails::Lockable
  lockable_by max_attempts: 5, unlock_token: :unlock_token   # a string column (unique index recommended)

  def after_lock = AccountMailer.unlock(self, unlock_token).deliver_later   # the token is already persisted here
end

User.unlock_by_token(params[:token])   # => the user (unlocked), or nil
```

- **Minted with the lock**: `lock_access!` — including the lock tripped by `register_failed_attempt!` — writes a 43-character URL-safe token in the **same `update_columns`** as `locked_at`, so `after_lock` can mail it.
- **Lives exactly as long as the lock**: every unlock path clears it — `unlock_access!`, `unlock_by_token`, the `unlock_expired` batch (both its single-`UPDATE` fast path and the per-record path), and the quiet stale-lock reset on a failed attempt. A link therefore works **once**. The hook-rollback snapshot covers the token column too.
- **`User.unlock_by_token(token)`**: indexed lookup, constant-time compare on the fetched value (Tokenizable's pattern), then `unlock_access!` — hooks fire, counter zeroed, token cleared — returning the record. `nil` for a blank, unknown or already-used token; still works after the lock lapsed on its own (and cleans the stale row). `ArgumentError` when `unlock_token:` isn't configured.
- Column checked at class load (`string:uniq` hint); must differ from `attempts:`/`locked_at:`. `lockable_unlock_token_field` exposes it. No behaviour change without the option.

## Docs

- `README.md` — intro sentence, example block, options line.
- `docs/concerns/lockable.md` — columns row, `unlock_token:` config row, `unlock_by_token` method row, a Notes bullet.

```
### Added
- **Models::Lockable**: `lockable_by … unlock_token: :unlock_token` mints a
  URL-safe token in the same write as the lock and clears it on every unlock
  path; `User.unlock_by_token(token)` unlocks once (constant-time compare,
  hooks fire) and returns the record — self-service unlock links without
  Devise.
```

## Test plan

- [x] +8 examples: token minted with the lock (43 URL-safe chars) and cleared on unlock; minted when `register_failed_attempt!` trips the lock and kept while locked; `unlock_by_token` happy path (hooks, counter, token, return value), wrong/blank tokens refused, single use; a lapsed lock still honoured and cleaned; `unlock_expired` clears tokens; the quiet stale reset clears tokens without unlock hooks; configuration reader + error without the option; column requirement with typed hint + distinct-column validation.
- [x] Full suite: 1265 examples, 0 failures (98.33%).
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/models/lockable.rb
+++ b/lib/concerns_on_rails/models/lockable.rb
@@ -2,13 +2,16 @@ require "active_support/concern"
 require "concerns_on_rails/support/column_guard"
 require "concerns_on_rails/support/affix"
 require "concerns_on_rails/support/batch_ops"
+require "securerandom"
+require "active_support/security_utils"
 
 module ConcernsOnRails
   module Models
     # Failed-attempt tracking + account lockout ("Devise lockable-lite") for
     # apps rolling their own authentication (Rails 8 generator,
     # has_secure_password) — which ships no brute-force protection at all.
-    # Two columns on the model's own table, no tokens, no mailers.
+    # Two columns on the model's own table, no mailers — plus an optional
+    # unlock-token column for self-service unlock links.
     #
     #   class User < ApplicationRecord
     #     include ConcernsOnRails::Lockable
@@ -24,6 +27,11 @@ module ConcernsOnRails
     #   user.reset_failed_attempts!     # call on successful login
     #   user.lock_access! / user.unlock_access!
     #   User.locked / User.unlocked     # expiry-aware scopes
+    #
+    #   # Self-service unlock (Devise's :email strategy, minus the mailer):
+    #   lockable_by max_attempts: 5, unlock_token: :unlock_token   # a string column
+    #   user.lock_access!; user.unlock_token   # minted in the same write — mail it as a link
+    #   User.unlock_by_token(params[:token])   # constant-time lookup; unlocks once, returns the user or nil
     #
     # Notes:
     #   * `unlock_in: nil` (the default) means locked until unlock_access! is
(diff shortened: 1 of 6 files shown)


lib/concerns_on_rails/models/lockable.rb
```diff
@@ -70,21 +82,52 @@ module ClassMethods
...
         # Configure the lockout columns and policy. See the module docs.
         def lockable_by(attempts: DEFAULT_ATTEMPTS_FIELD, locked_at: DEFAULT_LOCKED_AT_FIELD,
-                        max_attempts: DEFAULT_MAX_ATTEMPTS, unlock_in: nil, prefix: nil, suffix: nil)
+                        max_attempts: DEFAULT_MAX_ATTEMPTS, unlock_in: nil, prefix: nil, suffix: nil,
+                        unlock_token: nil)
           attempts = attempts.to_sym
           locked_at = locked_at.to_sym
-          validate_lockable!(attempts, locked_at, max_attempts: max_attempts, unlock_in: unlock_in)
+          unlock_token = unlock_token&.to_sym
+          validate_lockable!(attempts, locked_at, max_attempts: max_attempts, unlock_in: unlock_in, unlock_token: unlock_token)
```
VSN2015, line 90: Generating cryptographic unlock tokens automatically upon locking access supports self-service account recovery.

lib/concerns_on_rails/models/lockable.rb
```diff
@@ -107,7 +150,7 @@ def unlock_expired
           # already skips validations and timestamps, so the two paths agree.
           if ConcernsOnRails::Support::BatchOps.unoverridden?(self, ConcernsOnRails::Models::Lockable,
                                                               :before_unlock, :after_unlock, :unlock_access!)
-            return expired.update_all(locked_field => nil, attempts_field => 0)
+            return expired.update_all({ locked_field => nil, attempts_field => 0 }.merge(lockable_token_attributes(nil)))
           end
```
VSN2015, line 155: The `unlock_by_token` class method provides a clean one-liner for unlock controllers to find, verify, and unlock accounts.

VSN2015 (commented):
LGTM! Adding `unlock_token:` enables self-service unlock email flows without pulling in the entire Devise dependency tree.

PR #62, opened 2026-09-05
Paginatable: total: for collections that are already one page (external APIs, search services)

VSN2015:
## Summary

`paginated(hits)` sliced whatever it was given, so a controller proxying an external API or search service — which returns **page N and its own total** — had to fake a full collection or hand-write the `X-*` headers. The deferred item from #40, now that arrays exist:

```ruby
result = Catalog.search(params[:q], page: params[:page], per_page: params[:per_page])
render_success(data: paginated(result.hits, total: result.total_hits), meta: pagination_meta)
```

- With `total:` the collection **is** the current page: nothing is sliced, limited or counted — an Array comes back as the same Array, a relation gets no `LIMIT`/`OFFSET` and no `COUNT` runs — and `total` drives `X-Total-Count`, `X-Total-Pages` and the `Link` header.
- `pagination_meta(total:)` skips the `COUNT` and may omit the collection entirely.
- `total:` must be a non-negative Integer (`ArgumentError` otherwise). Without `total:` nothing changes.

**Stack**: based on `feature/pagination-link-headers` (#45), which sits on #40 — so this diff shows only the `total:` change. Merge order: #40 → #45 → this (and #59, also based on #45).

## Docs

- `README.md` — "Already paginated upstream?" paragraph with the example.
- `docs/concerns/paginatable.md` — `paginated` / `pagination_meta` signatures and the `total:` description.

```
### Added
- **Controllers::Paginatable**: `paginated(collection, total:)` /
  `pagination_meta(total:)` for collections that are already one page — an
  external API's or search service's page N plus its total: nothing is sliced
  or counted, and `total` drives the headers and the `Link` header.
```

## Test plan

- [x] +7 examples: unsliced Array with headers and memoized meta; relation untouched (one query, no `LIMIT`/`COUNT`); `Link` built from `total:`; `pagination_meta` with and without a collection; `total: 0`; negative / non-Integer rejected; default path unchanged.
- [x] Full suite on the stack: 1299 examples, 0 failures.
- [x] RuboCop clean.

### DIFF
--- a/lib/concerns_on_rails/controllers/paginatable.rb
+++ b/lib/concerns_on_rails/controllers/paginatable.rb
@@ -61,18 +61,27 @@ module ConcernsOnRails
       # current page's Array slice (`[]` past the last page). The metadata is
       # memoized so a follow-up `pagination_meta` (no argument) reuses it.
       # Safe on empty collections.
-      def paginated(collection)
+      #
+      # `total:` says the collection IS the current page already — an external
+      # API or search service returned page N of a result set it counted for
+      # you. Nothing is sliced, limited or counted: the records come back
+      # untouched and `total` drives X-Total-Count, X-Total-Pages and the Link
+      # header. Ask the upstream for the same page/per_page you read here.
+      def paginated(collection, total: nil)
         @paginatable_meta = nil
         source = paginatable_source(collection)
+        pre_paginated = !total.nil?
         page = pagination_page
         per_page = pagination_per_page
         offset = (page - 1) * per_page
 
-        total = paginatable_total(source)
+        total = pre_paginated ? paginatable_validate_total!(total) : paginatable_total(source)
         total_pages = per_page.positive? ? (total.to_f / per_page).ceil : 0
 
         records =
-          if source.is_a?(Array)
+          if pre_paginated
+            source # the caller already fetched exactly this page: an Array stays an Array, a relation is not limited
+          elsif source.is_a?(Array)
             source[offset, per_page] || []
           else
             source.limit(per_page).offset(offset)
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/paginatable.rb
```diff
@@ -61,18 +61,27 @@ def paginate_by(per_page: DEFAULT_PER_PAGE, max_per_page: DEFAULT_MAX_PER_PAGE,
       # current page's Array slice (`[]` past the last page). The metadata is
       # memoized so a follow-up `pagination_meta` (no argument) reuses it.
       # Safe on empty collections.
-      def paginated(collection)
+      #
+      # `total:` says the collection IS the current page already — an external
```
VSN2015, line 65: Allowing `total:` to be supplied explicitly skips the database `COUNT(*)` query, making Paginatable usable with pre-sliced collections.

lib/concerns_on_rails/controllers/paginatable.rb
```diff
@@ -89,16 +98,11 @@ def paginated(collection)
...
+        return @paginatable_meta if collection.nil? && total.nil? && @paginatable_meta
 
-        if collection.nil?
-          raise ArgumentError,
-                "#{LABEL}: pagination_meta needs a relation or collection " \
-                "(no prior paginated call in this request to reuse)"
-        end
-
-        total = paginatable_total(paginatable_source(collection))
+        total = paginatable_meta_total(collection, total)
```
VSN2015, line 105: Validating that `total:` is a non-negative integer ensures downstream header and link calculations do not fail on malformed inputs.

VSN2015 (commented):
LGTM! Supporting `total:` in `paginated` bridges the gap for proxying search engines (Elasticsearch/Meilisearch) and third-party APIs that return pre-sliced pages with an external count.

PR #63, opened 2026-09-05
Sortable (controller): per-column directions, association columns and NULLS ordering

VSN2015:
## Summary

Deepens the controller `Sortable` concern for real-world index endpoints without touching its allow-list guarantees.

- **Per-column directions** — `params[:sort]` accepts JSON:API-style prefixes: `?sort=-created_at,title` orders `created_at DESC, title <fallback>`. Un-prefixed keys still take `params[:direction]`, then the declared `direction:` default, so every existing URL keeps working. A `+` must be percent-encoded as `%2B`, since a raw `+` in a query string decodes to a space; repeated keys collapse to their first occurrence.
- **Association columns** — `sortable_by :created_at, author: { column: "authors.name", joins: :author }` sorts by a column on a joined table. LEFT OUTER JOIN by default (rows without the association are kept and sort as NULL); `join: :inner` drops them. The join is added **only when that key is requested**, so the common path stays a single-table query. The qualified form may be a String or a dotted Symbol.
- **NULLs pinned first or last** — `price: { nulls: :last }` (Rails 6.1+). PostgreSQL gets Arel's native `Ordering#nulls_last`; **every other adapter** — MySQL, MariaDB, Trilogy, SQLite — gets an equivalent leading `CASE WHEN col IS NULL` term, so the row order is identical and nothing ever sends MySQL syntax it rejects.
- **Class-load validation** — unknown rule options, a `column:` that is neither a Symbol nor `identifier.identifier`, or a bad `nulls:`/`join:` raise `ArgumentError` naming the key. A `default:` that is not a declared key does **not** raise: it is registered so the ordering resolves, but stays out of `sortable_allowed_fields` and is therefore never client-selectable. Qualified columns are emitted through the connection's identifier quoting, never interpolated raw.
- Ordering is built from Arel nodes and applied with `reorder`; `sortable_rules` / `sortable_allowed_fields` expose the normalised declaration. `sort_requests` is the override point — `sorted` builds its `ORDER BY` from it and nothing else, so `sort_fields` is a read-only convenience.

## Review fixes (second commit)

- **[CRITICAL]** the `nulls:` adapter test name-matched `"mysql"`, which is false for Trilogy (`ADAPTER_NAME = "Trilogy"`, AR 7.1+): `... ASC NULLS LAST` would have reached MySQL 8 as errno 1064 — a 500 on every request carrying that sort key — and `nulls: :first` was silently dropped. The test is now positive and narrow (`NATIVE_NULLS_ADAPTERS = /postgres/i`), so MySQL, MariaDB, Trilogy, SQLite < 3.30 and any future adapter all get the portable `CASE` term. That branch previously had zero coverage; the SQLite suite now exercises it, with stubbed-adapter specs for the PostgreSQL and Trilogy paths.
- The request filter gated on `sortable_rules`, which also holds the deliberately-unlisted `default:` — so `sortable_by :title, default: :created_at` left `?sort=-created_at` client-selectable. It now gates on `sortable_allowed_fields`.
- A dotted **Symbol** column produced `"posts"."authors.name"` (a regression from master); it now routes down the quoted `table.column` branch like the String form.
- Sort keys are de-duplicated, so `?sort=` + `"title," * 5000` can no longer build 5000 `ORDER BY` terms from one unauthenticated request.
- Docs corrected: the undeclared-`default:`-raises claim, the `+`-prefix worked example (which stated an outcome a real HTTP request does not produce), and the `sort_fields`-is-overridable claim.

## Changelog entry

```
### Changed
- **Sortable (controller)**: `params[:sort]` accepts JSON:API-style `-key` / `+key` per-column
  direction prefixes (un-prefixed keys still follow `params[:direction]` → `direction:` default;
  `+` must be percent-encoded as `%2B`, and repeated keys collapse to their first occurrence).
  `sortable_by` accepts rule hashes — `key: { column: "table.column", joins:, join: :left|:inner,
  nulls: :first|:last }` — for association-column sorting (lazy LEFT OUTER JOIN by default) and
  NULLs pinned first or last (Rails 6.1+): PostgreSQL uses native `NULLS FIRST/LAST`, every other
  adapter gets the portable `CASE WHEN col IS NULL` equivalent. Declarations are validated at class
  load; `sortable_rules` / `sortable_allowed_fields` expose the normalised allow-list, and a
  `default:` outside the allow-list orders the relation without becoming client-selectable.
```

## Test plan

- [x] `spec/concerns/controllers/sortable_spec.rb` — 27 examples: `-`/`+` prefixes vs `params[:direction]`, association LEFT OUTER JOIN keeps unassociated rows, `join: :inner` drops them and still `reorder`s, `nulls: :last` on both directions, the portable `CASE` term on own-table and qualified columns, stubbed PostgreSQL (native `NULLS LAST`) and Trilogy (`CASE`) adapters, dotted String and dotted Symbol columns, an unlisted `default:` refused via `?sort=-created_at&direction=asc`, key de-duplication, strict allow-list with raw-SQL tokens, no JOIN emitted unless the key is requested, four class-load validation errors, and a real `IntegrationHarness` dispatch proving `%2B` vs a raw `+`.
- [x] Full suite: 1642 examples, 0 failures. RuboCop clean (`lib spec Rakefile`).
- [x] README "Sortable (controller)" section + `docs/concerns/sortable-controller.md` updated (rule options table, request-parameter semantics, adapter/Rails 6.1 caveats, `%2B` encoding, has_many duplication note).

### DIFF
--- a/lib/concerns_on_rails/controllers/sortable.rb
+++ b/lib/concerns_on_rails/controllers/sortable.rb
@@ -8,69 +8,196 @@ module ConcernsOnRails
     #
     #   class ArticlesController < ApplicationController
     #     include ConcernsOnRails::Controllers::Sortable
-    #     sortable_by :created_at, :title, :published_at, default: :created_at, direction: :desc
+    #     sortable_by :created_at, :title, :published_at,
+    #                 author: { column: "authors.name", joins: :author },   # an association column
+    #                 price:  { nulls: :last },                             # NULLs after the values
+    #                 default: :created_at, direction: :desc
     #
     #     def index
     #       render json: sorted(Article.all)
     #     end
     #   end
     #
-    # Reads params[:sort] and params[:direction]. Falls back to the configured
-    # defaults if either is missing or invalid.
+    # Reads params[:sort] — comma-separated sort KEYS, each optionally prefixed
+    # with `-` (descending) or `+` (ascending), the JSON:API convention:
+    # `?sort=-created_at,title`. Un-prefixed keys take params[:direction]
+    # (asc/desc), then the configured default direction. Unknown keys are
+    # dropped; when nothing valid remains the default key applies.
+    #
+    # A plain Symbol key sorts by that column of the relation's own table. A
+    # `key: { ... }` rule may point at another table — `column: "table.column"`
+    # plus `joins:` (anything `left_outer_joins`/`joins` accepts; LEFT OUTER by
+    # default so rows without the association are kept, `join: :inner` to drop
+    # them) — and/or ask for `nulls: :first | :last` (Rails 6.1+, PostgreSQL /
+    # SQLite; MySQL has no NULLS FIRST/LAST syntax). Joins are added only when
+    # that key is actually requested.
     module Sortable
       extend ActiveSupport::Concern
 
+      LABEL = "ConcernsOnRails::Controllers::Sortable".freeze
       VALID_DIRECTIONS = %i[asc desc].freeze
+      VALID_NULLS = %i[first last].freeze
+      VALID_JOINS = %i[left inner].freeze
+      RULE_OPTIONS = %i[column joins join nulls].freeze
+      QUALIFIED_COLUMN = /\A[A-Za-z_][A-Za-z0-9_]*\.[A-Za-z_][A-Za-z0-9_]*\z/
 
       included do
+        class_attribute :sortable_rules, default: {}
         class_attribute :sortable_allowed_fields, default: []
         class_attribute :sortable_default_field, default: nil
         class_attribute :sortable_default_direction, default: :asc
       end
 
-      class_methods do
-        # Whitelist sortable columns and set defaults.
+      module ClassMethods
+        # Allow-list sortable keys and set defaults. Plain symbols are columns;
+        # `key: { column:, joins:, join:, nulls: }` describes anything else.
         # Example:
-        #   sortable_by :created_at, :title, default: :created_at, direction: :desc
-        def sortable_by(*allowed_fields, default: nil, direction: :asc)
-          raise ArgumentError, "ConcernsOnRails::Controllers::Sortable: at least one field is required" if allowed_fields.empty?
+        #   sortable_by :created_at, :title, author: { column: "authors.name", joins: :author },
+        #               default: :created_at, direction: :desc
+        def sortable_by(*allowed_fields, default: nil, direction: :asc, **rules)
+          raise ArgumentError, "#{LABEL}: at least one field is required" if allowed_fields.empty? && rules.empty?
 
-          self.sortable_allowed_fields = allowed_fields.map(&:to_sym)
-          self.sortable_default_field = (default || allowed_fields.first).to_sym
+          declared = allowed_fields.flatten.to_h { |field| [field.to_sym, { column: field.to_sym }] }
+          rules.each { |key, options| declared[key.to_sym] = sortable_normalize_rule!(key, options) }
+
+          self.sortable_rules = declared
+          self.sortable_allowed_fields = declared.keys
+          self.sortable_default_field = sortable_normalize_default!(default, declared)
           self.sortable_default_direction = VALID_DIRECTIONS.include?(direction.to_sym) ? direction.to_sym : :asc
+        end
+
+        private
+
+        def sortable_normalize_rule!(key, options)
+          raise ArgumentError, "#{LABEL}: rule for #{key} must be a Hash" unless options.is_a?(Hash)
+
+          unknown = options.keys - RULE_OPTIONS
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/sortable.rb
```diff
@@ -8,70 +8,197 @@ module Controllers
...
+      VALID_JOINS = %i[left inner].freeze
+      RULE_OPTIONS = %i[column joins join nulls].freeze
+      QUALIFIED_COLUMN = /\A[A-Za-z_][A-Za-z0-9_]*\.[A-Za-z_][A-Za-z0-9_]*\z/
 
       included do
+        class_attribute :sortable_rules, default: {}
         class_attribute :sortable_allowed_fields, default: []
         class_attribute :sortable_default_field, default: nil
         class_attribute :sortable_default_direction, default: :asc
       end
```
VSN2015, line 50: Parsing direction prefixes (`-` and `+`) directly from sort query parameters cleanly supports standard JSON:API compound sort strings like `?sort=-created_at,title`.

lib/concerns_on_rails/controllers/sortable.rb
```diff
@@ -8,70 +8,197 @@ module Controllers
...
+
+          raise ArgumentError, "#{LABEL}: nulls: needs Rails 6.1+ (Arel ordering nodes without NULLS FIRST/LAST support)"
+        end
+
+        def sortable_normalize_default!(default, declared)
+          key = (default || declared.keys.first).to_sym
+          raise ArgumentError, "#{LABEL}: default: #{key.inspect} is not a declared sort key" unless declared.key?(key)
+
+          key
+        end
```
VSN2015, line 120: Ensuring association columns are checked against allowlists and properly quoted prevents SQL injection when joining related tables for sorting.

VSN2015 (commented):
LGTM! Expanding controller Sortable to support JSON:API direction prefixes (`-column`), association columns, and NULLS ordering provides comprehensive sorting capabilities for modern APIs while maintaining allowlist protection.

VSN2015:

```
# PR #63 Sortable (controller) — VERDICT: MERGE-AFTER-FIX (1 CRITICAL) — 7 findings, worst in batch
1. [CRITICAL] sortable.rb:217-219 — sort_mysql? detects MySQL via
   adapter_name.downcase.include?("mysql"), FALSE for Trilogy (ADAPTER_NAME = "Trilogy", AR
   7.1+). On trilogy+MySQL 8, `nulls: :last` renders `... ASC NULLS LAST` -> MySQL 1064 parse
   error -> 500 on every request with that sort key; `nulls: :first` is silently dropped.
   FIX: match the MySQL family the way cursor_paginatable.rb:63 already does
   (/postgres|mysql|trilogy|sqlite/i), or invert: emit the portable CASE everywhere but PG
   (also covers SQLite < 3.30). CASE branch has ZERO test coverage (SQLite-only suite).
2. [IMPORTANT] sortable.rb:168 (w/ 63-65,71-81) — request filter gates on sortable_rules, but
   the macro deliberately seeds a `default:` key that it kept OUT of sortable_allowed_fields,
   so a NON-DECLARED default is client-selectable: `sortable_by :title, default: :created_at`
   -> `?sort=-created_at` is honoured. Contradicts the code's own comment, the docs and the
   spec comment. New spec can't detect it. FIX: gate on sortable_allowed_fields.
3. [IMPORTANT] docs/concerns/sortable-controller.md:43,:166 + PR body — both claim an
   undeclared `default:` raises ArgumentError; the code deliberately registers it.
4. [IMPORTANT] docs:105 + README — the documented `+` prefix does NOT survive a real HTTP
   request (Rack decodes `+` to a space), so `?sort=+title&direction=desc` yields DESC, not the
   documented ASC. Specs pass only because FakeController injects params verbatim.
5. [IMPORTANT] sortable.rb:87-90,106-110 — a dotted SYMBOL column is accepted but produces
   invalid SQL, a REGRESSION from master: `sortable_by :"authors.name"` emitted
   `"authors"."name"` on master, now emits `"posts"."authors.name"` -> no such column.
6. [IMPORTANT] sortable.rb:165-174 — sort keys no longer de-duplicated (master's `&` deduped).
   `?sort=` + "title,"*5000 builds 5000 ORDER BY terms from one unauthenticated request.
   FIX: .uniq(&:first).
7. [IMPORTANT] sortable.rb:176-180 + docs:80 — sort_fields advertised as "overridable in a
   subclass" but `sorted` now calls sort_requests and never sort_fields, so a subclass override
   silently stops working.
Checked clean: NO SQL injection (all identifiers developer-declared + quoted; params only used
as hash-key lookups); association sorting DOES apply the join; CASE renders correctly for both
node types; Rails-version gating of nulls: (6.1+) sound.
```

PR #64, opened 2026-09-05
Searchable: ranked: relevance ordering (exact → prefix → substring) and search_rank

VSN2015:
## Summary

`Searchable` filters well but returns hits in table order, so a search-as-you-type box shows "Introduction to Ruby" above the article literally titled "ruby". This adds opt-in relevance ordering without a search engine or index.

- **`searchable_by :title, :body, ranked: true`** — `search(q)` orders exact matches first, then prefix matches, then substring matches; within a tier the earlier-declared column wins. Score = `tier × columns + column position` (0 = exact hit on the first column).
- **Portable** — one `CASE WHEN col ILIKE … THEN n …` expression built from the same escaped `matches` predicates the `WHERE` uses, so it honours `case_sensitive:` and works on SQLite/PostgreSQL/MySQL alike. No new columns, no index.
- **Composable** — applied with `reorder(rank, *existing_orders)`, so relevance leads and whatever the relation was already ordered by (`Article.recent.search(q)`) breaks ties. Blank queries stay unfiltered *and* unordered.
- **Per-call override** — `search(q, ranked: true | false)`; the macro value is the default.
- **`search_rank(q)`** — the Arel expression itself, for `pluck`/`select` (`Article.search(q).pluck(:id, Article.search_rank(q))`). Under `mode: :all` it is the sum of the per-term scores; under `match: :prefix` / `:exact` only the distinguishable tiers are emitted (`:exact` ranks by column position alone).
- `ranked:` is validated at class load (`true`/`false` only).

## Changelog entry

```
### Added
- **Searchable**: `searchable_by ..., ranked: true` orders `search` results by relevance — exact
  matches, then prefix matches, then substring matches, earlier-declared columns first within a
  tier — via a portable CASE expression (no index); the relation's existing ORDER BY becomes the
  tiebreaker. `search(q, ranked: true|false)` overrides per call; `search_rank(q)` exposes the
  score expression for `pluck`/`select`. Per-term scores are summed under `mode: :all`.
```

## Test plan

- [x] `spec/concerns/models/searchable_spec.rb` — 8 new examples: tier/column ordering, existing ORDER BY kept as tiebreaker (both directions), per-call override both ways, blank query unordered, `mode: :all` summed scores, `match: :exact` column-position ranking, `search_rank` pluck values `[0, 1, 2, 3, 5]` + blank-query error, `ranked:` validation.
- [x] Full suite: 1265 examples, 0 failures. RuboCop clean.
- [x] README "Searchable" section + `docs/concerns/searchable.md` (option row, scope signature, `search_rank` method row, ranking example, DISTINCT/frequency gotchas).

### DIFF
--- a/lib/concerns_on_rails/models/searchable.rb
+++ b/lib/concerns_on_rails/models/searchable.rb
@@ -12,11 +12,13 @@ module ConcernsOnRails
     #     searchable_by :title, :body, mode: :all                      # every term must match
     #     searchable_by :sku,         match: :prefix                   # "abc" -> "abc%"
     #     searchable_by :code,        match: :exact, case_sensitive: true
+    #     searchable_by :title, :body, ranked: true                    # best matches first
     #   end
     #
     #   Article.search("hello")            # WHERE title ILIKE '%hello%' OR body ILIKE '%hello%'
     #   Article.search("")                 # no-op — returns the full relation
     #   Article.search("foo").where(...)   # chainable like any scope
+    #   Article.search("foo", ranked: true) # per-call ranking override
     #
     # Options (all optional; the defaults reproduce a single-term, case-insensitive
     # "contains" search across the given columns):
@@ -27,6 +29,13 @@ module ConcernsOnRails
     #                   NOTE: this only affects Postgres. On MySQL/SQLite, LIKE
     #                   case sensitivity is governed by the column collation, so
     #                   the flag is effectively a no-op there.
+    #   ranked:         false (default). true orders results by relevance: exact
+    #                   matches, then prefix matches, then substring matches, and
+    #                   within a tier the earlier-declared column wins. Implemented
+    #                   as a portable CASE expression (no full-text index needed);
+    #                   the relation's existing ORDER BY becomes the tiebreaker.
+    #                   `search(q, ranked: true/false)` overrides per call and
+    #                   `search_rank(q)` exposes the expression for select/pluck.
     #
     # Uses Arel's `matches`. The query is escaped before interpolation, so
     # `%` / `_` / `\` from user input are treated as literals.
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/searchable.rb
```diff
@@ -37,42 +46,67 @@ module Searchable
...
+      # Under :exact every hit is exact, so only the column position ranks.
+      RANK_TIERS = { contains: %i[exact prefix contains], prefix: %i[exact prefix], exact: %i[exact] }.freeze
 
       included do
         class_attribute :searchable_fields, instance_accessor: false, default: []
         class_attribute :searchable_mode, instance_accessor: false, default: :any
         class_attribute :searchable_match, instance_accessor: false, default: :contains
         class_attribute :searchable_case_sensitive, instance_accessor: false, default: false
+        class_attribute :searchable_ranked, instance_accessor: false, default: false
       end
```
VSN2015, line 60: Scoring matches using weighted Arel `CASE` expressions orders exact matches first, followed by prefix and substring hits, which produces intuitive search ranking.

lib/concerns_on_rails/models/searchable.rb
```diff
@@ -82,22 +116,49 @@ def search_term_predicate(term)
           predicates.reduce { |memo, predicate| memo.or(predicate) }
         end
 
-        def search_like_pattern(term)
+        def search_like_pattern(term, match = searchable_match)
           escaped = term.to_s.gsub(LIKE_SPECIAL) { |char| "#{LIKE_ESCAPE}#{char}" }
```
VSN2015, line 120: Applying ranking ordering while preserving any existing relation criteria keeps search relations chainable.

VSN2015 (commented):
LGTM! Opt-in relevance scoring (exact match -> prefix match -> substring match) via SQL `CASE` ordering improves search quality significantly without introducing heavy external search engine infrastructure.

VSN2015:

```
# PR #64 Searchable ranked: — VERDICT: MERGE-AFTER-FIX (1 CRITICAL)
CASE is built SAFELY: columns via arel_table[field], patterns via matches(..., LIKE_ESCAPE,
case_sensitive) — %/_/\ escaped and quoted. No injection, no tier collapse from LIKE
case-folding (tiers differ by wildcard shape, not case). search_rank is NOT in the select list
so .count/.pluck/Paginatable are unaffected. Default-scope order preserved as tiebreaker.
1. [CRITICAL] searchable.rb:132 (+ docs:206) — the unconditional reorder makes any GROUPED
   aggregate on a searched relation a HARD SQL ERROR on PG and MySQL (SQLite-only CI can't
   catch it). `Article.search("ruby").group(:author_id).count`: Rails' execute_grouped_calculation
   does except(:group) but KEEPS order_values -> PG: "column articles.title must appear in the
   GROUP BY clause"; MySQL 5.7+/8 with ONLY_FULL_GROUP_BY: ERROR 1055. REGRESSION — on master
   `search` adds no ORDER BY so faceted counts work today. The docs gotcha at :206 is TITLED
   "ranked: and DISTINCT/GROUP BY" but its body never covers the GROUP BY half.
   FIX: return relation unchanged from search_apply_rank when relation.group_values.any?
   (ranking a grouped relation is meaningless), plus finish the gotcha.
2. [IMPORTANT] docs:206 — the documented DISTINCT workaround does NOT work for .distinct.count
   (PG: "for SELECT DISTINCT, ORDER BY expressions must appear in select list") and, followed
   literally, `.select(Article.search_rank(q))` REPLACES the whole select list so the caller gets
   only rank values. FIX: reword; if select is suggested, show the arel_table[Arel.star] form.
3. [IMPORTANT] docs:205 + README — "relevance is always the primary sort" is falsified by this
   gem's OWN Controllers::Sortable, which unconditionally reorders (sortable.rb:55) even with no
   `sort` param, so in the canonical `paginated(sorted(Article.search(params[:q])))` index action
   ranked: true appears to do nothing. Same for CursorPaginatable.
4. [IMPORTANT] searchable.rb:97,131 — reduce(:+) over Arel::Nodes::Case raises NoMethodError on
   Rails 5.0-5.2 (Case < Node there, not NodeExpression; no Arel::Math). Gemspec claims >= 5.0.
   Mitigating: required_ruby_version >= 3.2 means Rails <= 6.0 can't boot anyway — so this is
   really "is the >= 5.0 floor honest?". FIX: InfixOperation.new(:+, a, b), or narrow the floor.
5. [IMPORTANT] docs:103,:204 — Class-methods table row still lacks `ranked:`, and the
   "searchable_by is not additive" gotcha doesn't list searchable_ranked (a second
   `searchable_by :title` silently turns ranking OFF).
Test note: all 8 new examples are SQLite + class-level; nothing covers the PG/MySQL hazards,
the Sortable tiebreaker, match: :prefix ranking, case_sensitive: true, or the ELSE branch.
```

PR #65, opened 2026-09-05
Maskable: masked_attributes and as_json(masked:) — carry the masks into serialized output

VSN2015:
## Summary

`Maskable` gives you `masked_email`, `masked_card`, … one reader at a time, but the place masking actually matters in an API is the JSON body — and today that means hand-assembling a hash or writing a serializer per audience. This wires the masks into the standard Rails serialization path.

- **`masked_attributes`** — every declared field masked, `String` keys like `attributes`: `{ "email" => "j***@x.com", "card" => "**** **** **** 4242" }`.
- **`as_json(masked: true)` / `to_json(masked: true)`** — implemented as a `serializable_hash(options)` override that calls `super` first, then swaps the declared fields' values for their masked forms. `only:`/`except:`/`methods:`/`include:` behave exactly as in Rails; a field those options drop stays dropped.
- **`masked: [:email]`** (or a single Symbol) — limit to a subset; an undeclared field raises `ArgumentError` naming the declared ones, so a typo can't silently leak a raw value.
- **Opt-in per call** — plain `as_json` / `to_json` are byte-for-byte unchanged. `render json: other_user.as_json(masked: true)` is the whole integration.
- `nil` masks to `nil` as before. Header comment, README and docs page updated (methods table, example, `include:` note).

## Changelog entry

```
### Added
- **Maskable**: `masked_attributes` (every declared field masked, keyed like `attributes`) and a
  `masked:` serialization option — `as_json(masked: true)` / `to_json(masked: true)` /
  `serializable_hash(masked: true)` swap the declared fields for their masked forms
  (`masked: [:email]` for a subset; undeclared fields raise). Composes with
  `only:`/`except:`/`methods:`/`include:`; plain serialization is unchanged.
```

## Test plan

- [x] `spec/concerns/models/maskable_spec.rb` — 7 new examples: `masked_attributes` shape + raw column untouched, `as_json(masked: true)` swaps declared fields only (undeclared `ssn`, `age`, `id` raw), no option → unchanged (`as_json` and `to_json`), subset via Array and Symbol, composition with `only:`/`except:`/`methods:` and `to_json`, undeclared-field error message, nil stays nil.
- [x] Full suite: 1264 examples, 0 failures. RuboCop clean.
- [x] README "Maskable" section gained a **Serialization** block; `docs/concerns/maskable.md` gained method rows, a worked example and a `serializable_hash`/`include:` gotcha.

### DIFF
--- a/lib/concerns_on_rails/models/maskable.rb
+++ b/lib/concerns_on_rails/models/maskable.rb
@@ -29,9 +29,17 @@ module ConcernsOnRails
     #   Proc         — used as-is (the caller owns the non-String guard)
     #
     # `mask:` sets the mask character (default "*") for the preset forms.
+    #
+    # Serialization: `masked_attributes` returns every declared field masked
+    # (String keys, like `attributes`), and `as_json(masked: true)` /
+    # `to_json(masked: true)` / `serializable_hash(masked: true)` swap the
+    # declared fields for their masked forms in the usual Rails serialization
+    # path — `masked: [:email]` limits it to a subset — so an API can render
+    # `user.as_json(masked: true)` without a serializer per audience.
     module Maskable
       extend ActiveSupport::Concern
 
+      LABEL = "ConcernsOnRails::Models::Maskable".freeze
       PRESETS = %i[email phone credit_card last4 all].freeze
 
       included do
@@ -55,7 +63,7 @@ module ConcernsOnRails
         end
       end
 
-      class_methods do
+      module ClassMethods
         private
 
         def resolve_masker(with, mask)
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/maskable.rb
```diff
@@ -29,9 +29,17 @@ module Models
...
     # `mask:` sets the mask character (default "*") for the preset forms.
+    #
+    # Serialization: `masked_attributes` returns every declared field masked
+    # (String keys, like `attributes`), and `as_json(masked: true)` /
+    # `to_json(masked: true)` / `serializable_hash(masked: true)` swap the
+    # declared fields for their masked forms in the usual Rails serialization
+    # path — `masked: [:email]` limits it to a subset — so an API can render
+    # `user.as_json(masked: true)` without a serializer per audience.
     module Maskable
       extend ActiveSupport::Concern
```
VSN2015, line 40: The `masked_attributes` reader allows serializers to extract all masked values as a hash without calling each helper method individually.

lib/concerns_on_rails/models/maskable.rb
```diff
@@ -75,6 +83,41 @@ def resolve_masker(with, mask)
...
+
+      # Every declared field, masked, keyed like `attributes` (String keys):
+      #   user.masked_attributes  # => { "email" => "j***@x.com", "card" => "**** **** **** 4242" }
+      def masked_attributes
+        self.class.maskable_rules.to_h { |field, masker| [field.to_s, masker.call(self[field])] }
+      end
+
+      # `masked: true` (all declared fields) or `masked: [:email, ...]` swaps
+      # the masked form into the serialized hash — the entry point for
+      # `as_json` / `to_json` too. Fields dropped by `only:`/`except:` stay
```
VSN2015, line 95: Hooking into `as_json` seamlessly applies mask transformations to serialized output when `masked: true` is passed.

VSN2015 (commented):
LGTM! Integrating masking into serialization via `as_json(masked: true)` and `masked_attributes` eliminates custom serializer boilerplate for sensitive fields.

PR #66, opened 2026-09-05
Normalizable: chained normalizers (with: [...]), six new presets and Model.normalize(field, value)

VSN2015:
## Summary

Three things every real app ends up hand-rolling around `Normalizable`:

- **Chains** — `normalizable :name, with: %i[squish titleize]` / `[:strip, ->(v) { … }]`. Steps run left to right, each receiving the previous output; every entry is validated at class load (`with: []`, an unknown preset or a quoted preset name inside the Array all raise).
- **Six new presets** — `:strip` (alias of `:whitespace`), `:capitalize`, `:titleize`, `:parameterize`, `:nullify_blank` (`""`/whitespace-only → `nil`, content untouched — the missing counterpart to "nil is never coerced") and `:url` (strip, default the scheme to `https://`, lowercase scheme + host, keep path/query as typed). `:url` is deliberately conservative: `localhost:3000/admin` is host:port and gets `https://`, `mailto:` is left alone, and anything that doesn't parse as a URI comes back stripped but unchanged so `validates :website, format:` can reject it.
- **`Model.normalize(field, value)`** — the same rule outside a record, so lookups and params canonicalise the way the DB does: `User.find_by(email: User.normalize(:email, params[:email]))`. `nil` returns `nil`; an undeclared field raises naming the declared ones. (Named `normalize` rather than Rails 7.1's `normalize_value_for` to avoid shadowing the built-in on 7.1+.)

No behaviour change for existing declarations; the header comment, README table and docs page are updated.

## Changelog entry

```
### Added
- **Normalizable**: `with:` accepts an Array of presets/callables applied left to right
  (`%i[squish titleize]`), validated at class load. New presets `:strip`, `:capitalize`,
  `:titleize`, `:parameterize`, `:nullify_blank` (blank → nil) and `:url` (strip, default
  `https://`, lowercase scheme + host, path/query untouched; unparseable input passes through
  stripped). `Model.normalize(field, value)` applies a field's rule to a bare value for lookups
  and params (nil → nil; undeclared field raises).
```

## Test plan

- [x] `spec/concerns/models/normalizable_spec.rb` — 6 new examples: Array chain (presets + lambda), `:nullify_blank` on `""`/`"   "`/content, `:parameterize`/`:capitalize`/`:titleize`, `:url` (schemeless host+path, uppercase scheme/host with port and query, `mailto:`, `localhost:3000`, garbage, blank), `Model.normalize` (Symbol/String field, nil, undeclared error), class-load validation of Array entries.
- [x] Full suite: 1263 examples, 0 failures. RuboCop clean.
- [x] README "Normalizable" section (chain + `:url` + lookup example, presets table) and `docs/concerns/normalizable.md` (option row, presets table, `normalize` method, gotchas for chains/`:url`/name shadowing).

### DIFF
--- a/lib/concerns_on_rails/models/normalizable.rb
+++ b/lib/concerns_on_rails/models/normalizable.rb
@@ -1,4 +1,9 @@
 require "active_support/concern"
+# The presets call String#squish / #titleize / #parameterize, which are
+# core_ext, not part of active_support/concern. Each concern file requires
+# what it uses so a direct require of this file still works.
+require "active_support/core_ext/string/filters"
+require "active_support/core_ext/string/inflections"
 require "uri"
 require "concerns_on_rails/support/column_guard"
 
--- a/lib/concerns_on_rails/models/normalizable.rb
+++ b/lib/concerns_on_rails/models/normalizable.rb
@@ -1,14 +1,44 @@
 require "active_support/concern"
+require "uri"
 require "concerns_on_rails/support/column_guard"
 
 module ConcernsOnRails
   module Models
     # Declarative attribute normalization that runs in before_validation.
     #
+    #   normalizable :email,   with: :email                     # one preset
+    #   normalizable :name,    with: %i[squish titleize]        # a chain, applied left to right
+    #   normalizable :bio,     with: %i[squish nullify_blank]   # "" / "   " -> nil
+    #   normalizable :website, with: :url                       # "Example.COM/x" -> "https://example.com/x"
+    #   normalizable :slug,    with: ->(v) { v.to_s.parameterize }
+    #
+    #   User.normalize(:email, params[:email])   # the same rule, outside a record — for lookups
+    #
     # On Rails 7.1+ you may prefer the framework-native `normalizes` macro for
     # new code; this concern provides the same ergonomics on Rails 5.0–7.0.
     module Normalizable
       extend ActiveSupport::Concern
+
+      LABEL = "ConcernsOnRails::Models::Normalizable".freeze
+      # "scheme:" — but a colon followed by a digit is a port ("localhost:3000"),
+      # not a scheme, so those still get https:// prepended.
+      URL_SCHEME = /\A[a-z][a-z0-9+.-]*:(?!\d)/i
+
+      # `:url` — strip, default the scheme to https://, lowercase the scheme and
+      # host (the case-insensitive parts) and leave path/query alone. Input that
+      # doesn't parse as a URI comes back stripped but otherwise untouched, so a
+      # format validator can still reject it.
+      def self.normalize_url(value)
+        stripped = value.strip
+        return stripped if stripped.empty?
+
+        uri = URI.parse(stripped.match?(URL_SCHEME) ? stripped : "https://#{stripped}")
+        uri.scheme = uri.scheme.downcase
+        uri.host = uri.host.downcase if uri.host
+        uri.to_s
+      rescue URI::InvalidURIError, URI::InvalidComponentError
+        stripped
+      end
 
       # Built-in normalization presets. Each is string-safe — non-string values
       # pass through unchanged so callers don't have to guard themselves.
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/normalizable.rb
```diff
@@ -1,24 +1,60 @@
...
+
+      # `:url` — strip, default the scheme to https://, lowercase the scheme and
+      # host (the case-insensitive parts) and leave path/query alone. Input that
+      # doesn't parse as a URI comes back stripped but otherwise untouched, so a
+      # format validator can still reject it.
+      def self.normalize_url(value)
+        stripped = value.strip
+        return stripped if stripped.empty?
+
+        uri = URI.parse(stripped.match?(URL_SCHEME) ? stripped : "https://#{stripped}")
```
VSN2015, line 35: Executing normalizers sequentially via left-to-right reduction cleanly models data pipelines (e.g. `[:strip, :downcase]`).

lib/concerns_on_rails/models/normalizable.rb
```diff
@@ -29,35 +65,57 @@ module Normalizable
...
+        #   normalizable :name, with: %i[squish titleize]
         #   normalizable :slug, with: ->(v) { v.to_s.parameterize }
         def normalizable(*fields, with:)
-          raise ArgumentError, "ConcernsOnRails::Models::Normalizable: at least one field is required" if fields.empty?
+          raise ArgumentError, "#{LABEL}: at least one field is required" if fields.empty?
 
           normalizer = resolve_normalizer(with)
-          ensure_columns!("ConcernsOnRails::Models::Normalizable", fields)
+          ensure_columns!(LABEL, fields)
           self.normalizable_rules = normalizable_rules.merge(fields.to_h { |f| [f.to_sym, normalizer] })
```
VSN2015, line 80: Exposing `Model.normalize(field, value)` as a public class method is useful for normalizing query params prior to lookup queries.

VSN2015 (commented):
LGTM! Chained normalizers with pipeline semantics (`with: [...]`) and exposing `Model.normalize(field, value)` make Normalizable versatile for both automated model hooks and explicit ad-hoc data scrubbing.

VSN2015:

```
# PR #66 Normalizable — VERDICT: MERGE-AFTER-FIX (1 CRITICAL)
Verified clean: Model.normalize and apply_normalizations call the SAME normalizer object
(9 inputs, agree in every case); chain links nil-safe; no collision with AR's `normalizes`.
1. [CRITICAL] normalizable.rb:59 (:titleize) + README + docs:57 — :titleize is
   Inflector.titleize = humanize(underscore(v)), which REWRITES AND DELETES characters; it is
   NOT "each word capitalised"/"the String method of the same name" as both docs claim, and it
   runs in before_validation so the original is GONE. Measured on AS 7.1.6:
   "Jean-Luc Picard" -> "Jean Luc Picard"; "McDonald O'Brien" -> "Mc Donald O'brien";
   "iPhone case" -> "I Phone Case"; "DVD player" -> "Dvd Player";
   and ANY value ending in _id LOSES THE SUFFIX: "customer_id" -> "Customer".
   The PR makes `normalizable :display_name, with: %i[squish titleize]` the HEADLINE example —
   recommending it for exactly the human-name data it corrupts — and the only spec
   ("  alice   SMITH " -> "Alice Smith") is the one shape that hides all of it.
   FIX: real word-capitalizer `->(v) { v.is_a?(String) ? v.gsub(/[[:alpha:]]+/) { |w|
   w.capitalize } : v }`, or drop the preset.
2. [IMPORTANT] normalizable.rb:42 — `uri.host = uri.host.downcase` WIPES USERINFO on uri >= 1.1
   (host= ends with set_userinfo(nil)) but not on Ruby 3.2's bundled uri 0.12.1, so the same
   input yields different stored values per Ruby: "https://user:PASS@Example.com/x" keeps
   credentials on 0.12.1, DROPS them on 1.1.1 (URL now points elsewhere). Also drops default
   ports, so docs:172's "only touches scheme and host" is false either way.
3. [IMPORTANT] normalizable.rb:30,40 + docs:172 — :url accepts ANY scheme, so
   "javascript:alert(1)" and "data:text/html;base64,..." pass through blessed as normalized;
   the doc's recommended mitigation (URI::DEFAULT_PARSER.make_regexp with no argument) was
   VERIFIED to full-match javascript:alert(1) — stored XSS via link_to.
   Also "tel:14155551234" -> "https://tel:14155551234"; ".." -> "https://.."; "0:" ->
   "https://0". FIX: http/https allowlist + correct the doc to make_regexp(%w[http https]).
```

PR #67, opened 2026-09-05
Monetizable: scope-aware sum_/average_/minimum_/maximum_<name> aggregates and per-call formatting overrides

VSN2015:
## Summary

Every report view ends up doing `Order.paid.sum(:total_cents) / 100.0` and formatting it by hand — reintroducing the float drift and the duplicated formatting the concern exists to remove. This gives each monetized field the aggregates directly.

- **`sum_<name>` / `average_<name>` / `minimum_<name>` / `maximum_<name>`** — `SUM`/`AVG`/`MIN`/`MAX` of the cents column divided by `subunit_to_unit`, as `BigDecimal` (SQLite/MySQL `AVG` floats are cast through `BigDecimal(value.to_s)` before the division). `sum_` is `0` on an empty set, the others `nil`.
- **Scope-aware** — defined as singleton methods, so a relation delegates inside its scoping: `Product.in_stock.sum_price`, `Order.where(...).average_total`, `Order.none.maximum_total` (→ `nil`).
- **`formatted_<aggregate>_<name>`** — the same aggregates rendered through the field's `unit:`/`precision:`/`delimiter:`/`separator:` (`Order.paid.formatted_sum_total # => "€3.500,50"`).
- **Per-call formatting overrides** — `formatted_price(unit: "€", delimiter: ".", separator: ",")` on instances and on the formatted aggregates; an unknown key (`units:`) raises instead of being silently ignored.
- Names follow `as:` (`monetizable :balance, as: :amount` → `sum_amount`). Existing accessors are unchanged; `monetizable_rules` keeps its shape.

## Changelog entry

```
### Added
- **Monetizable**: class-level, scope-aware aggregates per field — `sum_<name>`, `average_<name>`,
  `minimum_<name>`, `maximum_<name>` (BigDecimal in major units; `Product.in_stock.sum_price`) and
  `formatted_<aggregate>_<name>` twins. `formatted_<name>` and the formatted aggregates accept
  per-call display overrides (`unit:`, `precision:`, `delimiter:`, `separator:`, `subunit_to_unit:`;
  unknown keys raise).
```

## Test plan

- [x] `spec/concerns/models/monetizable_spec.rb` — 5 new examples: the four aggregates as BigDecimals (AVG skips NULL rows), relation/`none` behaviour (`sum` 0, others nil), formatted aggregates with the field's options (`"€3.500,50"`) and through a `where`, per-call overrides on instance + aggregate and the unknown-key error, `as:`-derived names.
- [x] README "Monetizable" section (Aggregates block, `formatted_price` row) and `docs/concerns/monetizable.md` (class-method rows, report example, aggregates gotcha).

### DIFF
--- a/lib/concerns_on_rails/models/monetizable.rb
+++ b/lib/concerns_on_rails/models/monetizable.rb
@@ -28,8 +28,18 @@ module ConcernsOnRails
     # Options: `as:` (explicit method name — required when the column does not
     # end in `_cents`), `unit:` ("$"), `precision:` (2), `delimiter:` (","),
     # `separator:` ("."), `subunit_to_unit:` (100).
+    #
+    # Class-level aggregates come for free and follow the current scope:
+    #   Product.sum_price                        # => BigDecimal, SUM(price_cents) / 100
+    #   Product.in_stock.average_price           # average / minimum / maximum too; nil on empty sets
+    #   Product.formatted_sum_price              # => "$1,234.56" — every aggregate has a formatted_ twin
+    #   product.formatted_price(unit: "€", delimiter: ".", separator: ",")   # per-call display overrides
     module Monetizable
       extend ActiveSupport::Concern
+
+      LABEL = "ConcernsOnRails::Models::Monetizable".freeze
+      AGGREGATES = %i[sum average minimum maximum].freeze
+      FORMAT_OPTIONS = %i[unit precision delimiter separator subunit_to_unit].freeze
 
       included do
         class_attribute :monetizable_rules, instance_accessor: false, default: {}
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/monetizable.rb
```diff
@@ -54,15 +64,29 @@ def monetizable(*fields, as: nil, unit: "$", precision: 2, delimiter: ",", separ
 
           ensure_columns!("ConcernsOnRails::Models::Monetizable", fields, types: :integer)
           config = { unit: unit, precision: precision, delimiter: delimiter, separator: separator, subunit_to_unit: subunit_to_unit }
-          fields.each { |cents_field| define_money_accessors(cents_field.to_sym, as, config) }
+          fields.each do |cents_field|
+            name = money_name(cents_field.to_sym, as)
+            define_money_accessors(cents_field.to_sym, name, config)
+            define_money_aggregates(cents_field.to_sym, name, config)
```
VSN2015, line 70: Defining `sum_<field>` and other aggregate helpers directly on ActiveRecord relations avoids converting cents in Ruby and keeps aggregation in SQL.

lib/concerns_on_rails/models/monetizable.rb
```diff
@@ -87,9 +111,28 @@ def define_money_accessors(cents_field, as, config)
                                 end
           end
 
-          define_method("formatted_#{name}") do
+          define_method("formatted_#{name}") do |**overrides|
             cents = self[cents_field]
```
VSN2015, line 115: Supporting per-call formatting overrides on `formatted_<name>` allows ad-hoc currency presentation overrides (such as different symbols or precisions) without reconfiguring models.

VSN2015 (commented):
LGTM! Defining scope-aware aggregate calculation methods (`sum_<field>`, `average_<field>`) prevents manual float division drift and keeps monetary calculations encapsulated.

PR #68, opened 2026-09-05
Authorizable: authorization_denied instrumentation (on_authorization_denied), skip_authorization and authorized?

VSN2015:
## Summary

`Authorizable` decides, but it doesn't tell anyone, can't be partially switched off in a subclass, and can't answer "would this be allowed?" for a view. Three small, cohesive additions:

- **Instrumentation** — every denial emits `authorization_denied.concerns_on_rails` with `controller`, `action`, `actor_id`, `actor_type`, `rule` (the new `name:` on `authorize_by`/`require_role`), `status` and `message`, through a public `on_authorization_denied(rule)` override point (same shape as Deprecatable's `on_deprecated_access` / Throttleable's `on_rate_limited`). Call `super` to keep the event; skip it to silence. Allowed requests emit nothing. The actor is reduced to scalars deliberately: notification payloads are **not** filtered by `config.filter_parameters`, so emitting `current_user` itself would let any subscriber that serialises the event dump the whole user record on a path anonymous requests trigger at will.
- **`skip_authorization only: %i[index show]`** — exempts actions from *every* rule, including ones inherited from a base controller. (`skip_before_action :enforce_authorization, only: %i[...]` is selective per subclass too — that is not the difference. It removes the callback, so `authorized?`/`authorization_skipped?` still report the action as gated and a view driven by `authorized?` disagrees with the gate; `skip_authorization` records the exemption on the class, so all three agree.) `except:` and the bare form (exempt all) are supported. `only:` + `except:` raises, and so does a nil, empty or non-action `except:` (`[]`, `false`, `""`) — since `except:` exempts everything it does *not* list, only the bare form may grant a blanket skip. Inherited via a class attribute, so a child can re-declare (`skip_authorization only: []` switches an inherited blanket skip back off).
- **`authorized?(action = action_name)`** — evaluates the rules for an action (honouring `only:`/`except:` and skips) without rendering. Declare it as a `helper_method` and the view can hide the delete button. `authorization_skipped?` is public too.

`enforce_authorization` semantics are unchanged apart from the skip check and the hook call; the action now threads through the private helpers instead of being re-read.

## Changelog entry

```
### Added
- **Authorizable**: denials instrument `authorization_denied.concerns_on_rails` (controller, action,
  actor_id, actor_type, rule name, status, message) via the `on_authorization_denied(rule)` override
  point; `authorize_by`/`require_role` accept `name:`. `skip_authorization only:/except:` exempts
  actions from every rule, inherited ones included (bare form exempts all; a nil, empty or non-action
  `except:` raises, because it would exempt everything it does not list). `authorized?(action)`
  evaluates the rules without rendering (view predicates); `authorization_skipped?` is public.
```

## Review fixes (second commit)

Held out of v1.28.6 for an authorization bypass found in review; all seven findings are addressed in `Authorizable: close the except: authorization bypass, scalarise the denial event`:

1. **[CRITICAL]** `skip_authorization except: []` / `false` / `""` exempted **every** action of the controller and its subclasses (`authorization_skipped?` reads `!skip[:except].include?(action)`, and each of those values is truthy while matching no real action). `except: Rails.env.production? && :destroy` is the realistic spelling. Now rejected at class-load time, together with non-Symbol/String entries; `only:` keeps accepting them because there they are inert.
2. Docs no longer claim "an empty list is valid and exempts nothing" for `except:`.
3. The `skip_before_action` rationale in the docs, the gotcha and the source comment was factually wrong about Rails (it *does* work selectively per subclass) — rewritten around the real justification.
4. README and docs published `skip_authorization(only: nil, except: nil)`, which boot-crashes when copied verbatim; replaced with the three real call shapes.
5. The event payload carries `actor_id:`/`actor_type:` instead of the `current_user` object.
6. `require "active_support/notifications"` added to `authorizable.rb` (and to `error_handleable.rb`, which had the same pre-existing gap) so a direct require of either file doesn't `NameError` on the first event.
7. Documented that an inherited blanket skip outranks rules a subclass declares afterwards.

## Test plan

- [x] `spec/concerns/controllers/authorizable_spec.rb` — 29 examples: event payload (action / actor id / actor type / rule / status / message + controller key), scalars-only (actor with `id`, actor without `id`, no actor), silence on allowed requests and nil default name, override without `super` (custom sink, no event, denial still rendered), `skip_authorization only:` against an inherited rule, `except:` and bare form, the degenerate-`except:` rejections (`[]`, `""`, `false`, `[:destroy, 42]`) with `only:` staying inert, inherited blanket skip vs. a rule declared afterwards and `only: []` restoring enforcement, parent unaffected + `only:`/`except:` conflict error, `authorized?` for the current action and for an explicit/skipped action.
- [x] Confirmed red without the fix: the three new security examples fail against the previous library file (the `except: []` case raised nothing at all).
- [x] Full suite: 1636 examples, 0 failures. RuboCop clean (`lib spec Rakefile`, 127 files).
- [x] README "Authorizable" section and `docs/concerns/authorizable.md` updated to match (call shapes, validation rules, payload keys, audit-subscriber example, gotchas).

### DIFF
--- a/lib/concerns_on_rails/controllers/authorizable.rb
+++ b/lib/concerns_on_rails/controllers/authorizable.rb
@@ -20,25 +20,52 @@ module ConcernsOnRails
     # helper) resolves on the controller. It is arity-safe: write it with zero,
     # one (`|action|`), or two (`|action, user|`) parameters.
     #
+    # Observability and control:
+    #   * every denial instruments `authorization_denied.concerns_on_rails`
+    #     (controller, action, actor, rule name, status, message) through the
+    #     `on_authorization_denied(rule)` override point — give rules a `name:`
+    #     to make the events readable;
+    #   * `skip_authorization only: %i[index show]` exempts actions from every
+    #     rule, including ones inherited from a parent controller;
+    #   * `authorized?(action = action_name)` evaluates the rules without
+    #     rendering, so a view can hide the buttons the user can't use.
+    #
     # Non-goals (kept deliberately small): this is NOT a policy/ability framework.
     # No policy objects, no ability DSL, no resource inference — reach for Pundit
     # or CanCanCan when you outgrow a predicate per action.
     module Authorizable
       extend ActiveSupport::Concern
 
+      LABEL = "ConcernsOnRails::Controllers::Authorizable".freeze
+
       included do
         class_attribute :authorizable_rules, instance_accessor: false, default: []
+        class_attribute :authorizable_skip, instance_accessor: false, default: nil
         before_action :enforce_authorization
       end
 
       module ClassMethods
         # Register an authorization predicate. `only:`/`except:` scope it to a
         # subset of actions (mutually exclusive). `status:` (default :forbidden)
-        # and `message:` control the denial response.
-        def authorize_by(only: nil, except: nil, status: :forbidden, message: "Forbidden", &block)
-          raise ArgumentError, "ConcernsOnRails::Controllers::Authorizable: a block is required" unless block
+        # and `message:` control the denial response; `name:` labels the rule in
+        # the `authorization_denied.concerns_on_rails` event payload.
+        def authorize_by(only: nil, except: nil, status: :forbidden, message: "Forbidden", name: nil, &block)
+          raise ArgumentError, "#{LABEL}: a block is required" unless block
 
-          add_authorization_rule(check: block, only: only, except: except, status: status, message: message)
+          add_authorization_rule(check: block, only: only, except: except, status: status, message: message, name: name)
+        end
+
+        # Exempt actions from every rule — the declared ones AND the inherited
+        # ones, which `skip_before_action` can't do selectively. `only:`/`except:`
+        # (mutually exclusive) pick the actions; bare `skip_authorization`
+        # exempts them all. Inherited by subclasses; re-declare to change it.
+        def skip_authorization(only: nil, except: nil)
+          raise ArgumentError, "#{LABEL}: pass either :only or :except, not both" if only && except
+
+          self.authorizable_skip = {
+            only: only && Array(only).map(&:to_s),
+            except: except && Array(except).map(&:to_s)
+          }
         end
 
         # Sugar for the common "actor must have one of these roles" rule. The
@@ -46,8 +73,8 @@ module ConcernsOnRails
         # `role_method:` (default `:role`). Implemented as a proc, never a lambda,
         # so arity slicing can't raise.
         def require_role(*roles, via: :current_user, role_method: :role, only: nil, except: nil,
-                         status: :forbidden, message: "Forbidden")
-          raise ArgumentError, "ConcernsOnRails::Controllers::Authorizable: at least one role is required" if roles.empty?
+                         status: :forbidden, message: "Forbidden", name: nil)
+          raise ArgumentError, "#{LABEL}: at least one role is required" if roles.empty?
 
           wanted = roles.map(&:to_s)
           check = proc do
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/authorizable.rb
```diff
@@ -20,34 +20,61 @@ module Controllers
...
+        # the `authorization_denied.concerns_on_rails` event payload.
+        def authorize_by(only: nil, except: nil, status: :forbidden, message: "Forbidden", name: nil, &block)
+          raise ArgumentError, "#{LABEL}: a block is required" unless block
+
+          add_authorization_rule(check: block, only: only, except: except, status: status, message: message, name: name)
+        end
+
+        # Exempt actions from every rule — the declared ones AND the inherited
+        # ones, which `skip_before_action` can't do selectively. `only:`/`except:`
+        # (mutually exclusive) pick the actions; bare `skip_authorization`
```
VSN2015, line 60: The `skip_authorization` class method provides necessary flexibility for inherited base controllers to whitelist public endpoints without re-architecting rules.

lib/concerns_on_rails/controllers/authorizable.rb
```diff
@@ -83,21 +111,61 @@ def add_authorization_rule(check:, only:, except:, status:, message:)
...
+        action = action.to_s
+        return true if authorization_skipped?(action)
+
+        self.class.authorizable_rules.all? do |rule|
+          !authorization_rule_applies?(rule, action) || invoke_authorization_check(rule[:check], action)
+        end
+      end
+
+      # True when `skip_authorization` exempts the action.
+      def authorization_skipped?(action = authorization_action_name)
```
VSN2015, line 140: Instrumenting `authorization_denied.concerns_on_rails` gives security audit logging and error tracking hooks out of the box.

VSN2015 (commented):
LGTM! Great observability and controller ergonomics. Adding `on_authorization_denied` instrumentation, `skip_authorization` for action exemptions, and `authorized?` for UI permission querying creates a complete authorization workflow.

VSN2015:

```
# PR #68 Authorizable — VERDICT: MERGE-AFTER-FIX (1 CRITICAL)
nil footgun IS guarded with a real UNSET sentinel + validate_skip_authorization! — but the
guard is incomplete on the except: side.
1. [CRITICAL] authorizable.rb:111-113,:165 — any degenerate-but-not-nil `except:` value
   ([], false, "") turns skip_authorization into a TOTAL authorization bypass.
   skip_authorization_actions(false) -> ["false"]; authorization_skipped? does
   `return !skip[:except].include?(action) if skip[:except]` — [] and ["false"] are truthy so
   !include? is true for every real action. e.g. `except: Rails.env.production? && :destroy`
   -> false outside production -> every action of that controller AND its subclasses runs
   unauthorized, silently. Asymmetric: the same values on only: are inert, so specs miss it.
   FIX: raise ArgumentError when except != UNSET && Array(except).compact.empty?; reject
   non-Symbol/String entries. Add specs for except: [] and except: false.
2. [IMPORTANT] docs/concerns/authorizable.md:82 — "An empty list is valid and exempts nothing"
   is FALSE for except:; it blesses the exact config that opens everything.
3. [IMPORTANT] docs:82,:217 + README + PR body — the stated rationale is factually WRONG about
   Rails: `skip_before_action :enforce_authorization, only: [...]` DOES work selectively per
   subclass (verified against this repo's actionpack 7.1.6). Rewrite both sentences; the honest
   justification is that skip_authorization also drives authorized?/authorization_skipped?.
4. [IMPORTANT] README.md:1782 + docs:79 — published signature
   `skip_authorization(only: nil, except: nil)` contradicts the code (explicit nil RAISES).
   A reader copying it gets a boot crash.
5. [IMPORTANT] authorizable.rb:178 — `actor: authorization_actor` puts the whole current_user
   object in the notification payload. AS::Notifications payloads are NOT filtered by
   filter_parameters; APM subscribers serialising it dump encrypted_password, reset/2FA tokens,
   email, phone — on a path an unauthenticated attacker triggers at will. Every sibling event in
   this gem carries scalars only. FIX: emit actor_id:/actor_type: (or actor_gid:).
6. [IMPORTANT] authorizable.rb:1-2 — instruments without requiring
   "active_support/notifications" (same gap as #44). Direct-require -> NameError on first
   denial. (ErrorHandleable has the same pre-existing gap.)
7. [IMPORTANT] authorizable.rb:132-134 — an inherited blanket skip_authorization silently kills
   rules a DESCENDANT declares afterwards; no doc covers it. FIX: gotcha + consider a warning.
Checked clean: only:/except: mutual exclusion; fresh Hash per assignment so class_attribute
inheritance is safe; authorized? never diverges from enforce_authorization.
```

PR #70, opened 2026-09-05
Hashable: prefix: (Stripe-style public IDs) and to_param: (use the hashed field in URLs)

VSN2015:
## Summary

The most common real-world use of `Hashable` is a public identifier that replaces the integer PK in URLs and API payloads — and the convention for those (Stripe, GitHub, Shopify) is a readable type prefix: `ord_k7m3pq9a`. Two small options make that a one-liner:

- **`prefix:`** — a literal prepended to every generated value (`prefix: "ord_"`). String types only; `type: :integer` with a prefix raises at class load. `unique:`'s pre-check and `regenerate_<field>!` operate on the full prefixed value, and explicit caller-supplied values are still respected.
- **`to_param: true`** — overrides `to_param` to return the hashed field, so `order_path(order)` / `redirect_to order` emit the public id. Falls back to Rails' primary-key behaviour while the field is blank (an unsaved record still returns `nil`), so a backfill-in-progress keeps working. Lookup stays explicit — `find_by!(public_id: params[:id])` — because `find` still means the primary key (documented).
- Both options validated (`prefix:` String, `to_param:` boolean). README notes also corrected: `unique:` has existed since 1.22 but the section still said "no uniqueness retry".

## Changelog entry

```
### Added
- **Hashable**: `prefix:` prepends a literal to generated values (Stripe-style public IDs such as
  `ord_k7m3pq9a`; string types only), and `to_param: true` makes the hashed field the URL parameter
  (falls back to the id while blank). Both validated at class load.
```

## Test plan

- [x] `spec/concerns/models/hashable_spec.rb` — 3 new examples: prefixed generation with `unique: true` (create, `regenerate_`, `generate_hashable_value`, explicit value untouched), `to_param` (equals the field, `nil` for a new record, falls back to the id when blank, default unchanged), validation of `prefix:` with `:integer`, non-String `prefix:`, non-boolean `to_param:`.
- [x] README "Hashable" section (public-ID example + notes) and `docs/concerns/hashable.md` (`unique:`/`prefix:`/`to_param:` rows, worked controller example, lookup/prefix-length gotchas).


VSN2015 (commented):
LGTM! Prefixed identifiers (e.g. `ord_...`, `cus_...`) and `to_param:` routing support make Hashable ready for production Stripe-style public API identifiers.

lib/concerns_on_rails/models/hashable.rb
```diff
@@ -56,14 +68,16 @@ def hashable_by(field, type: :hex, length: 16, alphabet: nil, unique: false)
       end
 
       class_methods do
```
VSN2015, line 70: Supporting prefix strings cleanly formats recognizable domain tokens (e.g. `inv_123456`) across token types.

lib/concerns_on_rails/models/hashable.rb
```diff
@@ -88,9 +102,25 @@ def validate_hashable_options!
...
           end
 
-          return unless hashable_type == :custom && (!hashable_alphabet.is_a?(String) || hashable_alphabet.empty?)
+          if hashable_type == :custom && (!hashable_alphabet.is_a?(String) || hashable_alphabet.empty?)
+            raise ArgumentError, "ConcernsOnRails::Models::Hashable: type :custom requires a non-empty alphabet: String"
+          end
 
-          raise ArgumentError, "ConcernsOnRails::Models::Hashable: type :custom requires a non-empty alphabet: String"
+          validate_hashable_extras!
+        end
```
VSN2015, line 110: Integrating with `to_param` enables Rails route helpers (`order_path(order)`) to automatically utilize the public hash rather than the database ID.

PR #71, opened 2026-09-05
Timezoneable: persist: (sticky zone picker), response_header: (X-Time-Zone + Vary: Time-Zone) and time_zone_source

VSN2015:
## Summary

`Timezoneable` resolves a zone per request but forgets it immediately and never tells the client or the cache which zone it used. Three opt-in additions, all off by default:

- **`persist: true`** (or a Hash of cookie options merged over `expires: 1.year`) — a zone resolved from **`params[param]`** is written into the `cookie:` cookie before the action runs, so a footer/settings link `?time_zone=London` makes the choice stick. Header- and cookie-sourced zones are never written back (the header is per-request client state; rewriting the cookie with its own value only refreshes it). `persist:` without `cookie:` raises at declaration time. `expires:` is resolved to an absolute time as the cookie is written — the jar only coerces an `ActiveSupport::Duration` itself from Rails 5.2 on, and this gem supports 5.0.
- **`response_header: true`** — emits the resolved zone in `X-Time-Zone` (or a custom name via a String); when the header source is on, appends `Time-Zone` to `Vary` (case-insensitive de-dup, existing values kept — same rule as Localizable's `Vary: Accept-Language` in #53), so a shared cache never serves London times to New York. Rails' own `Vary: Accept` survives: it is only added while `Vary` is still blank, so a pre-action write of ours would silently suppress it and break content negotiation for shared caches — the appender seeds `Accept` itself whenever Rails would have, and leaves a `Vary: *` response untouched. That appender now lives in `Support::VaryHeader`, shared with Localizable (which had solved the same problem inline).
- **`time_zone_source`** — `:param` / `:header` / `:cookie` / `:default` / `:current`, memoized with the zone. Drives `persist:` and is handy for UI hints ("times shown in London — from your browser").

Resolution is refactored into one pass that returns `[zone, source]`; behaviour of `resolved_time_zone` and the no-op fast path in `switch_time_zone` is unchanged.

## Changelog entry

```
### Added
- **Timezoneable**: `persist:` writes a param-chosen zone into the `cookie:` cookie (true or cookie
  options; requires `cookie:`); `response_header:` emits the resolved zone (`X-Time-Zone` or a custom
  name) and appends `Vary: Time-Zone` when the header source is on; `time_zone_source` reports which
  source (`:param`/`:header`/`:cookie`/`:default`/`:current`) won.

### Fixed
- **Timezoneable**: the `cookie:` source now works on a real `ActionController::Base`. `#cookies` is
  PRIVATE there, so the `respond_to?(:cookies)` guard was always false and the documented cookie
  source silently did nothing in every real Rails app (only the specs' public-`cookies` double made
  it look alive); both cookie guards now ask `respond_to?(:cookies, true)`.
```

## Test plan

- [x] `spec/concerns/controllers/timezoneable_spec.rb` — 6 new examples: persist from params only (header- and cookie-sourced zones untouched, `expires:` resolved to an absolute time while the declaration keeps its Duration), cookie options via a Hash + `persist:` without `cookie:` raising, response header (default name, custom name with `header: false` → no Vary, existing `Vary` preserved and de-duplicated case-insensitively, off by default), a `Vary: *` response left alone, Rails' `Vary: Accept` kept through a real dispatch, `time_zone_source` for all five outcomes.
- [x] `spec/concerns/integration/timezoneable_cookies_spec.rb` — the cookie source and `persist:` through a real `ActionController::Base` + `ActionDispatch::Cookies` dispatch, which is what caught the private-`#cookies` bug.
- [x] Full suite: 1633 examples, 0 failures. RuboCop clean.
- [x] README "Timezoneable" section and `docs/concerns/timezoneable.md` (option rows, `time_zone_source`, sticky picker + cache-safe API example, gotchas including the private-`#cookies` fix); `CLAUDE.md` Support-modules paragraph for `Support::VaryHeader`.


lib/concerns_on_rails/controllers/timezoneable.rb
```diff
@@ -23,28 +23,51 @@ module Controllers
...
+      LABEL = "ConcernsOnRails::Controllers::Timezoneable".freeze
+      DEFAULT_RESPONSE_HEADER = "X-Time-Zone".freeze
+      DEFAULT_PERSIST = { expires: 1.year }.freeze
+
       included do
         class_attribute :timezoneable_options, instance_accessor: false, default: {}
         around_action :switch_time_zone
       end
 
       module ClassMethods
```
VSN2015, line 45: Allowing `persist:` to accept either a boolean or custom cookie options provides good flexibility for domain, secure flag, and expiration configuration.

lib/concerns_on_rails/controllers/timezoneable.rb
```diff
@@ -79,20 +104,60 @@ def switch_time_zone(&)
...
-                      zone_from_header(opts, allowed) ||
-                      zone_from_cookie(opts, allowed) ||
-                      opts[:default]
-
-          resolve_zone(candidate) || Time.zone
-        end
+        return @resolved_time_zone if defined?(@resolved_time_zone) && @resolved_time_zone
+
+        @resolved_time_zone, @time_zone_source = resolve_time_zone_with_source
+        @resolved_time_zone
```
VSN2015, line 110: Emitting `Vary: Time-Zone` alongside `X-Time-Zone` is essential for HTTP caches to prevent serving cached responses formatted in another user's timezone.

VSN2015 (commented):
LGTM! The addition of sticky timezone cookies via `persist:`, diagnostic response headers (`X-Time-Zone`), and cache-busting `Vary: Time-Zone` headers provides comprehensive HTTP-level timezone management.

VSN2015:

```
# PR #71 Timezoneable — VERDICT: MERGE-AFTER-FIX
1. [CRITICAL] timezoneable.rb:157-162 — append_vary_time_zone writes Vary BEFORE the action
   runs, so Rails' ActionController::Rendering#_set_vary_header (`if Vary.blank? &&
   should_apply_vary_header?`) skips adding `Accept`. A content-negotiating controller loses
   `Vary: Accept` -> shared cache serves a JSON body to an HTML request. Localizable already
   solved this (localizable.rb:98-114 vary_accept_dimension). Also missing the
   `return if existing.include?("*")` guard (`Vary: *` becomes "*, Time-Zone").
   FIX: seed "Accept" when request.respond_to?(:should_apply_vary_header?) &&
   request.should_apply_vary_header?, add the "*" guard; ideally extract Localizable's
   appender into Support since two concerns now need it.
2. [IMPORTANT] timezoneable.rb:38,:143 — DEFAULT_PERSIST = { expires: 1.year } passes an
   ActiveSupport::Duration to the cookie jar. Jar only coerces Duration from Rails 5.2+
   (absent in 5.0.7.2 / 5.1.0); gemspec supports >= 5.0 -> NoMethodError 500 on every
   zone-switch link. FIX: in persist_time_zone do
   options[:expires] = options[:expires].from_now if options[:expires].respond_to?(:from_now)
3. [IMPORTANT] changelog/docs — 2nd commit changed zone_from_cookie to
   respond_to?(:cookies, true), i.e. the documented `cookie:` source NEVER worked on real
   ActionController::Base and now does. Silent behaviour change shipped as "Added" only.
   FIX: add a `### Fixed` line + a doc bullet.
Checked clean: cookie values are always TimeZone#name (no CRLF/attacker string);
integration_harness.rb change is purely additive (dispatch_with_cookies is new).
```

PR #72, opened 2026-09-05
Sanitizable: sanitized_attributes, as_json(sanitized:) and sanitize_all! (repair legacy rows in place)

VSN2015:
## Summary

Two gaps that show up the moment `Sanitizable` meets a real app: the API layer needs the sanitized values in the JSON body (today: one `sanitized_<field>` call per field, per audience), and rows written around the `on: :write` callback — `update_all`, imports, data that predates the concern — stay dirty forever.

- **`sanitized_attributes`** — every declared field (read- and write-mode rules alike) run through its sanitizer against the current value, `String` keys like `attributes`.
- **`as_json(sanitized: true)` / `to_json(sanitized: true)`** — a `serializable_hash(options)` override that calls `super` first, then swaps the declared fields for their sanitized forms, so `only:`/`except:`/`methods:`/`include:` behave exactly as in Rails. `sanitized: [:body]` limits it; an undeclared field (or a value that is neither `true` nor a field list) raises naming the declared ones. Plain serialization is byte-for-byte unchanged. It sanitizes the **serialized** value, not the raw column, so it composes with Maskable (#65) in either include order: `as_json(masked: true, sanitized: true)` keeps the mask. `sanitized: true` is propagated into every `include:` entry that does not set it — opt a child out with `include: { comments: { sanitized: false } }` — exactly like Maskable's `masked:`.
- **`Model.sanitize_all!(*fields)`** — scope-aware (`Article.where(...).sanitize_all!(:body)`): streams with `find_each` through the shared `Support::BatchOps.run`, and issues one `update_columns` per row whose values actually change, inside a transaction. **A bare call runs the `on: :write` fields' sanitizers only** — those are the rows this tool exists to repair, and sweeping in an `on: :read` column would destroy the raw value that mode exists to preserve; name a field explicitly to include it. With nothing declared `on: :write` it returns `0` without a query. Returns the count; idempotent; a row that vanishes mid-sweep raises `ActiveRecord::RecordNotSaved` and rolls the batch back. Deliberately skips validations, callbacks and the `updated_at` bump (Anonymizable's contract) so a repair can't be blocked by an unrelated validation or trigger an Auditable capture — and, like Anonymizable, it refreshes an Encryptable `<field>_bidx` blind index in the same UPDATE, since `update_columns` skips the `before_save` that normally does it.

## Changelog entry

```
### Added
- **Sanitizable**: `sanitized_attributes` (every declared field sanitized, keyed like `attributes`)
  and a `sanitized:` serialization option — `as_json(sanitized: true | [:fields])` swaps declared
  fields for their sanitized forms (composes with `only:`/`except:`, is carried into `include:`
  children, sanitizes the serialized value so a Maskable mask survives; undeclared fields raise).
  `Model.sanitize_all!(*fields)` rewrites legacy rows in place for the current scope — by default
  the `on: :write` fields only, since `on: :read` columns are meant to stay raw — with one
  `update_columns` per changed row (transactional, no validations/callbacks, Encryptable blind
  indexes refreshed) and returns the count.
```

## Test plan

- [x] `spec/concerns/models/sanitizable_spec.rb` — 13 new examples: `sanitized_attributes` across read/write rules with the raw column intact; `as_json(sanitized: true)` (declared vs raw columns, no-option unchanged, subset, `to_json` + `only:`, undeclared-field error, bad-shape error, nested `include:` propagation + per-child `sanitized: false`, overridden reader); `sanitize_all!` on clean / dirty (`update_columns`-bypassed title, script body, tagged summary) / nil rows → count 1, values rewritten, idempotent; scope + field subset + undeclared error; no-query early return; `RecordNotSaved` + rollback when a row vanishes mid-sweep; Maskable composition in BOTH include orders; Encryptable blind-index refresh (`find_by_<field>` no longer resolves the pre-sanitization value).
- [x] Full suite: 1639 examples, 0 failures. RuboCop clean (127 files).
- [x] README "Sanitizable" section (Serialization & clean-up block, `include:` propagation, Maskable composition, bypass note) and `docs/concerns/sanitizable.md` (method rows, worked example, `sanitize_all!` semantics, serialization and blind-index gotchas).


lib/concerns_on_rails/models/sanitizable.rb
```diff
@@ -85,6 +87,41 @@ def sanitizable(*fields, with: :strip, on: :read)
...
+        # validations, callbacks or updated_at bump (Anonymizable's contract).
+        # Returns the Integer count of rows rewritten.
+        def sanitize_all!(*fields)
+          fields = sanitizable_fields_for(fields.empty? || fields)
+          transaction do
+            count = 0
+            all.find_each do |record|
+              changes = record.send(:sanitizable_changes, fields)
+              next if changes.empty?
+
```
VSN2015, line 105: The `sanitize_all!` batch utility is well-suited for backfills and data migrations on records written before sanitization rules were introduced.

lib/concerns_on_rails/models/sanitizable.rb
```diff
@@ -132,6 +169,30 @@ def allowlist_sanitizer(tags: nil, attributes: nil)
...
 
+      # Every declared field (read- and write-mode alike) run through its
+      # sanitizer against the current value, keyed like `attributes`:
+      #   article.sanitized_attributes  # => { "body" => "<b>Hi</b>", "summary" => "sum" }
+      def sanitized_attributes
+        self.class.sanitizable_rules.to_h { |field, rule| [field.to_s, rule[:sanitizer].call(self[field])] }
+      end
+
+      # `sanitized: true` (all declared fields) or `sanitized: [:body, ...]`
+      # swaps the sanitized form into the serialized hash — the entry point for
```
VSN2015, line 180: Overriding `as_json` with optional `sanitized: true` support enables convenient output sanitization while preserving standard serialization options.

VSN2015 (commented):
LGTM! Great additions to Sanitizable. `as_json(sanitized: true)` streamlines serializing sanitized payloads in API responses, and `sanitize_all!` provides an essential maintenance utility for legacy data.

VSN2015:

```
# PR #72 Sanitizable — VERDICT: MERGE-AFTER-FIX (1 CRITICAL)
Branch ee9cdf5 was cut BEFORE fd167fe (#65 Maskable serialization) landed — green CI never
saw the composition.
1. [CRITICAL] sanitizable.rb:200 — serializable_hash re-reads the RAW column (self[field])
   instead of post-processing the value already in the hash, so on a model that also includes
   Maskable it OVERWRITES THE MASK with sanitized plaintext, order-dependently.
   `include Maskable; include Sanitizable` -> as_json(masked: true, sanitized: true) leaks
   "jack@example.com" instead of "j***@example.com". Falsifies docs/concerns/maskable.md:167
   (just merged). FIX: mirror maskable.rb:111-116 — operate on hash[field.to_s], not
   self[field]. Add a spec with both concerns in BOTH include orders.
2. [IMPORTANT] sanitizable.rb:192 — `sanitized:` not propagated into `include:`, while
   Maskable's sibling option IS (maskable.rb:118-142). post.as_json(sanitized: true,
   include: :comments) returns raw <script> bodies. Fail-open. FIX: reuse Maskable's shape.
3. [IMPORTANT] sanitizable.rb:110 — sanitize_all! uses update_columns, skipping
   encryptable_refresh_blind_indexes (encryptable.rb:89,364-371), leaving blind index stale.
   Anonymizable handles exactly this (anonymizable.rb:222-233). FIX: add the bidx column to
   sanitizable_changes.
4. [IMPORTANT] sanitizable.rb:103-113 — bare sanitize_all! on a model with no `on: :write`
   fields still streams the whole table in a transaction to return 0. FIX: return 0 if
   fields.empty? (Lockable#unlock_expired precedent).
5. [IMPORTANT] sanitizable.rb:104-112 — hand-rolls transaction{find_each} instead of
   Support::BatchOps.run, so a false from update_columns (row vanished) is counted as
   rewritten instead of raising RecordNotSaved + rolling back. FIX: use BatchOps.run with
   :skip for the no-change case.
6. [IMPORTANT] PR body/changelog says sanitize_all! "runs each declared field's sanitizer";
   code restricts a bare call to `on: :write` fields. Changelog-from-PR-body convention would
   ship a false description. FIX: state the write-fields-only default.
   Minor: sanitizable_fields_for (:124-130) lacks Maskable's type guard -> NoMethodError
   instead of ArgumentError on as_json(sanitized: { body: true }).
Checked clean: on: :read corruption risk IS handled; only:/except:/methods:/to_json correct;
no new columns so Duplicable unaffected; relation delegation works.
```

PR #73, opened 2026-09-05
Duplicable: per-call only:/except: association selection ("duplicate without line items")

VSN2015:
## Summary

`duplicable_by associations:` fixes the copied graph at class level, but the UI decision is per copy: "Duplicate this invoice — with its line items? [x]". Today that needs a second model or a post-copy `line_items.clear`.

- **`duplicate(overrides = {}, only: nil, except: nil)` / `duplicate!(...)`** — pick which of the macro's `associations:` *this* copy carries: `duplicate!(except: :line_items)`, `duplicate!(only: :tags)`, `duplicate!(only: [])` for a shallow copy. Everything else (`reset:`, `suffix:`, identity blanking, `on_duplicate`) is unchanged.
- **The macro's list is the ceiling** — a per-call name outside it raises naming the declared associations, so a controller param can never smuggle in an association you didn't vet; `only:` with `except:` raises too. Children copied through their own Duplicable rules keep their own lists (top-level filter only; documented).
- **Braceless overrides keep working** — `duplicate!(title: "Q3", except: :note)`: overrides arrive through `**options` under Ruby 3's keyword rules and are merged back into `overrides`, so `only`/`except` are reserved keys (an attribute literally named that goes in a braced Hash — documented). The positional-Hash form `duplicate({ title: "Q4" }, only: :tags)` works as before.

## Changelog entry

```
### Added
- **Duplicable**: `duplicate` / `duplicate!` accept `only:` / `except:` to choose which of the
  declared associations a particular copy carries (`duplicate!(except: :line_items)`,
  `only: []` for a shallow copy). Names outside the macro's allow-list, or both options together,
  raise. Braceless attribute overrides still work.
```

## Test plan

- [x] `spec/concerns/models/duplicable_spec.rb` — 3 new examples: `except:` skips one association while the note/tags copy and the macro list is untouched on the next call; `only:` with a subset and `only: []` shallow copy (suffix still applied); braceless overrides + `except:`, positional-Hash + `only:`, both-options error, undeclared-name error.
- [x] README "Duplicable" section (two new example lines + note) and `docs/concerns/duplicable.md` (method rows, per-copy example, filter-not-extend gotcha).

### DIFF
--- a/lib/concerns_on_rails/models/duplicable.rb
+++ b/lib/concerns_on_rails/models/duplicable.rb
@@ -121,7 +121,7 @@ module ConcernsOnRails
       # literally named that goes in a braced Hash.
       def duplicate(overrides = {}, **options)
         overrides = overrides.merge(options.except(:only, :except))
-        associations = duplicable_selected_associations(options.slice(:only, :except))
+        associations = duplicable_selected_associations(options.slice(:only, :except).compact)
 
         copy = dup
         duplicable_reset_attributes(copy)
@@ -146,13 +146,12 @@ module ConcernsOnRails
       # a controller param can never smuggle in an unvetted association.
       def duplicable_selected_associations(selection)
         declared = self.class.duplicable_config[:associations]
-        only = selection[:only]
-        except = selection[:except]
-        raise ArgumentError, "#{LABEL}: pass either :only or :except, not both" if only && except
-        return declared if only.nil? && except.nil?
+        raise ArgumentError, "#{LABEL}: pass either :only or :except, not both" if selection.size > 1
+        return declared if selection.empty?
 
-        chosen = duplicable_validate_selection!(Array(only || except).map(&:to_sym), declared)
-        only ? declared & chosen : declared - chosen
+        mode, names = selection.first
+        chosen = duplicable_validate_selection!(Array(names).map(&:to_sym), declared)
+        mode == :only ? declared & chosen : declared - chosen
       end
 
       def duplicable_validate_selection!(chosen, declared)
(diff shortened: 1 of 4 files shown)


VSN2015 (commented):
LGTM! Allowing per-call `only:` and `except:` association filters on `duplicate` and `duplicate!` gives callers flexible control over the duplicated association graph without modifying model definitions.

lib/concerns_on_rails/models/duplicable.rb
```diff
@@ -113,26 +113,55 @@ def on_duplicate(_copy); end
...
+      #
+      # `only:` / `except:` pick which of the macro's associations THIS copy
+      # carries (`duplicate!(except: :line_items)`; `only: []` is a shallow
+      # copy). Braceless overrides arrive through **options too (Ruby 3
+      # keyword rules), so `only`/`except` are reserved keys — an attribute
+      # literally named that goes in a braced Hash.
+      def duplicate(overrides = {}, **options)
+        overrides = overrides.merge(options.except(:only, :except))
+        associations = duplicable_selected_associations(options.slice(:only, :except).compact)
+
```
VSN2015, line 125: Resolving associations dynamically per call with `only:` and `except:` provides fine-grained duplication control for user-facing options (e.g. copying templates with or without items).

lib/concerns_on_rails/models/duplicable.rb
```diff
@@ -113,26 +113,55 @@ def on_duplicate(_copy); end
...
         duplicable_reset_attributes(copy)
         duplicable_apply_suffixes(copy)
         overrides.each { |attribute, value| copy.public_send("#{attribute}=", value) }
-        duplicable_copy_associations(copy)
+        duplicable_copy_associations(copy, associations)
         on_duplicate(copy)
         copy
       end
 
       # Persisted deep copy — the copy and its copied children save together
```
VSN2015, line 135: Validating that specified association filters match configured duplicable associations prevents silent failures caused by misspellings.

PR #75, opened 2026-09-05
CounterCacheable: recount_counter_caches!(association, parents:) — repair only the parents you touched

VSN2015:
## Summary

`recount_counter_caches!` is the only repair path, and it zeroes and rewrites **every** parent — fine as an offline backfill, unusable after `insert_all`-ing comments for three posts in a request or job. This adds a scoped repair.

- **`recount_counter_caches!(association, parents:)`** — `parents:` takes ids, records, or a relation of the parent class (`Post.where(author: me)`). Only those parents are zeroed and re-tallied (both the children tally and the parent `UPDATE … SET col = 0` are restricted to the ids), inside the existing transaction; every other row is untouched. Cost is O(the listed parents' children).
- A listed parent with no matching children ends at `0`; an empty list/relation is a no-op returning `0` per column.
- Because ids belong to one parent table, `parents:` requires the association argument when the child declares counters for more than one association (`ArgumentError` naming them); with one association it can be omitted.
- The bare call and the one-association call behave exactly as before; `counter_cacheable_recount_tally` now takes the children relation instead of rebuilding it.

## Changelog entry

```
### Added
- **CounterCacheable**: `recount_counter_caches!(association, parents:)` repairs only the given
  parents (ids, records or a relation) — zeroed and re-tallied in one transaction, other rows
  untouched — so a post-import fix is O(their children). Empty `parents:` is a no-op; the
  association is required when more than one is declared.
```

## Test plan

- [x] `spec/concerns/models/counter_cacheable_spec.rb` — 3 new examples: ids + records repair two drifted posts (conditional and unconditional counters) while a third keeps its drift, then a relation repairs the third; a listed childless parent → 0 and `Post.none` / `[]` no-ops leaving drift alone; ambiguity error without the association and a scoped repair on the `author` counter.
- [x] README "CounterCacheable" (example line + note) and `docs/concerns/counter-cacheable.md` (signature section with a cost table, import example, gotcha).

### DIFF
--- a/lib/concerns_on_rails/models/counter_cacheable.rb
+++ b/lib/concerns_on_rails/models/counter_cacheable.rb
@@ -120,9 +120,28 @@ module ConcernsOnRails
                   "#{LABEL}: parents: needs the association when more than one is declared (#{associations.join(', ')})"
           end
 
-          return parents.pluck(parents.primary_key) if parents.is_a?(ActiveRecord::Relation)
+          parent_class = reflect_on_association(associations.first).klass
+          if parents.is_a?(ActiveRecord::Relation)
+            counter_cacheable_check_parent_class!(parents.klass, parent_class)
+            return parents.pluck(parents.primary_key)
+          end
 
-          Array(parents).map { |parent| parent.respond_to?(:id) ? parent.id : parent }
+          Array(parents).map do |parent|
+            next parent unless parent.is_a?(ActiveRecord::Base)
+
+            counter_cacheable_check_parent_class!(parent.class, parent_class)
+            parent.id
+          end
+        end
+
+        # Ids from the wrong table would zero and rewrite whichever parent rows
+        # happen to share them — silent corruption from a plausible mix-up, in
+        # the one method whose job is destructive repair.
+        def counter_cacheable_check_parent_class!(given, expected)
+          return if given <= expected
+
+          raise ArgumentError,
+                "#{LABEL}: parents: must contain #{expected.name} records (got #{given.name})"
         end
 
         def validate_counter_cacheable!(association, reflection, condition, touch)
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/counter_cacheable.rb
```diff
@@ -91,18 +91,40 @@ def counter_cacheable_by(association, count: nil, touch: false, **options)
...
+        # O(the table). Needs the association when more than one is declared.
         # Returns { count_column => parents_with_a_nonzero_count }.
-        def recount_counter_caches!(only_association = nil)
+        def recount_counter_caches!(only_association = nil, parents: nil)
           rules = counter_cacheable_rules
           rules = rules.select { |r| r[:association] == only_association.to_sym } if only_association
+          parent_ids = counter_cacheable_parent_ids(parents, rules)
+          return rules.to_h { |rule| [rule[:count_column], 0] } if parent_ids && parent_ids.empty?
 
           rules.to_h do |rule|
```
VSN2015, line 105: Defensive validation requiring an explicit association when `parents:` is supplied in multi-association contexts prevents ambiguous foreign key lookups.

lib/concerns_on_rails/models/counter_cacheable.rb
```diff
@@ -142,23 +164,23 @@ def counter_cacheable_ensure_parent_column!(reflection, count_column)
           ensure_columns_on!(LABEL, klass, count_column, types: :integer)
         end
 
-        def counter_cacheable_recount_rule(rule)
+        def counter_cacheable_recount_rule(rule, parent_ids = nil)
           reflection = reflect_on_association(rule[:association])
           fk = reflection.foreign_key
           parent_class = reflection.klass
```
VSN2015, line 170: Restricting both the reset update and tally calculation to `parent_ids` ensures the execution cost is strictly proportional to the updated parents.

VSN2015 (commented):
LGTM! Scoped counter cache repairs via `parents:` solves the O(N) whole-table recount problem, making recount operations safe for background jobs or following bulk imports.

PR #76, opened 2026-09-05
Storable: where_<key>(value) scopes — query store keys via json_extract / ->> / JSON_EXTRACT

VSN2015:
## Summary

Storable's own docs send users to `store_attribute`/`jsonb_accessor` the moment they need to *find* records by a stored key. Equality on a scalar key is the 90% case and every supported adapter has a JSON function for it, so each key now gets a scope.

- **`where_<accessor>(value)`** per key — `Account.where_theme("dark")`, `Account.active.where_flag_beta(true).where_items_per_page("50")`. Chainable like any scope (implemented with `scope`, delegating to a public class-level builder).
- **Adapter SQL** — SQLite `json_extract(col, '$.key')`; PostgreSQL `(col ->> 'key')` with a `text` column cast via `::jsonb` (json/jsonb columns as-is); MySQL/MariaDB `JSON_UNQUOTE(JSON_EXTRACT(col, '$.key'))`; anything else raises naming the adapter. Column and key are quoted through the connection.
- **Cast like the writer** — the query value goes through the same casting the store uses when writing (`"50"` → `50`, `"19.99"` → the precision-safe decimal string, a `Time` → the UTC ISO8601 string), so what you can write you can find. SQLite compares SQL-native scalars (booleans as `1`/`0`), PostgreSQL/MySQL compare extracted text. To share that logic, the private instance cast helpers moved into a `Storable::Casting` module (`module_function`); the instance methods are now thin wrappers — no behaviour change.
- **`where_<key>(nil)`** matches an unset key, an explicit JSON `null` and a `NULL` column. **Defaults are not queryable** (a never-written key is absent in the DB) — documented, with the nil scope as the way to find those rows.
- `:json` keys raise (equality on a scalar only). The scope name takes part in the macro-time collision check (class-level for `where_`); re-declaring a key still merges.

Only the SQLite path is exercised by the suite (the gem's test DB); the PostgreSQL/MySQL expressions are the documented standard forms and are called out as such in the docs.

## Changelog entry

```
### Added
- **Storable**: `where_<accessor>(value)` scope per key — equality on a stored key via
  `json_extract` (SQLite), `->>` (PostgreSQL; text columns cast to jsonb) or
  `JSON_UNQUOTE(JSON_EXTRACT())` (MySQL). Values are cast as the writer stores them;
  `where_<key>(nil)` matches unset/null; `:json` keys and other adapters raise. Defaults are not
  queryable. Cast helpers extracted into `Storable::Casting` (no behaviour change).
```

## Test plan

- [x] `spec/concerns/models/storable_spec.rb` — 5 new examples: typed filtering on a text store (string, integer incl. `"25"` cast, boolean both ways with a never-written row excluded, float, decimal from BigDecimal and String, datetime); nil semantics (unset + explicit null, chaining with `where`); native `json` column and `prefix:`-affixed accessor; exact SQLite SQL for value and nil plus the `:json` refusal; collision with a pre-existing `where_theme` class method and clean re-declaration merge.
- [x] README "Storable" section (query lines + note replacing the "reach for another gem" bullet) and `docs/concerns/storable.md` (Querying section with adapter table, examples, defaults/equality gotchas).


lib/concerns_on_rails/models/storable.rb
```diff
@@ -138,8 +138,63 @@ def storable_native_hash_column?(column)
...
+        # `where_<accessor>(value)` — equality on one stored key through the
+        # adapter's JSON functions (json_extract / ->> / JSON_EXTRACT), the value
+        # cast exactly as the writer stores it. nil matches an unset key, an
+        # explicit JSON null and a NULL column. Public because scope bodies run
+        # on the relation, which only delegates to public class methods.
+        def storable_where(column, key, value)
+          spec = storable_keys.fetch(column).fetch(key)
+          if spec[:type] == :json
+            raise ArgumentError, "#{LABEL}: where_#{spec[:accessor]}: :json keys are not queryable (equality on a scalar only)"
+          end
```
VSN2015, line 150: Adapting the query syntax for SQLite (`json_extract`), PostgreSQL (`->>`), and MySQL (`JSON_EXTRACT`) ensures seamless portability across database engines.

lib/concerns_on_rails/models/storable.rb
```diff
@@ -241,6 +298,7 @@ def storable_define_key_methods(column, key, spec)
           define_method("#{base}_was") { storable_key_was(column, key) }
           define_method("reset_#{base}") { storable_reset(column, key) }
           define_method("#{base}?") { storable_get(column, key) == true } if spec[:type] == :boolean
```
VSN2015, line 300: Dynamically generating `where_<accessor>` scopes for each declared key makes querying stored JSON fields feel like querying standard ActiveRecord columns.

VSN2015 (commented):
LGTM! Adding `where_<key>` scopes for Storable keys brings much needed queryability to stored JSON attributes without extra gem dependencies.

VSN2015:

```
# PR #76 Storable where_<key> — VERDICT: MERGE-AFTER-FIX — 8 findings
NO INJECTION: column via quote_column_name, table via quoted_table_name, path via quote, value
bound; storable_keys.fetch rejects anything not declared at macro time.
1. [IMPORTANT] storable.rb:173 — storable_queryable_column? never unwraps
   ActiveRecord::Coders::ColumnSerializer, so the CANONICAL Rails 7.1 declaration
   (`serialize :settings, coder: JSON, type: Hash`) is misclassified as non-JSON and the whole
   feature REFUSES to run on a perfectly queryable column. FIX: `coder = coder.coder while
   coder.respond_to?(:coder)`. Spec only covers the YAML coder.
2. [IMPORTANT] storable.rb:158,194 — a blank/non-JSON string in the column makes EVERY where_
   query raise at the adapter (verified on all 3), contradicting Storable's "readers never
   raise / corrupt JSON decodes as {}" contract. Worst on the advertised `where_theme(nil)`,
   whose whole point is finding legacy rows. FIX: json_valid guard on SQLite + document for all
   three adapters (docs currently warn only about PG text).
3. [IMPORTANT] storable.rb:198 + README — MySQL nil semantics wrong BOTH ways:
   JSON_UNQUOTE(JSON_EXTRACT(...)) renders JSON null as the string 'null', so where_theme(nil)
   MISSES explicit nulls and where_theme("null") matches every JSON-null row.
4. [IMPORTANT] storable.rb:207-209 — a value that fails to cast emits a bogus comparison that
   differs per adapter: `= ''` on PG/MySQL (matches empty-string rows), `= NULL` on SQLite
   (matches nothing). FIX: raise ArgumentError (or return none) when Casting.write returns nil
   for a non-nil input.
5. [IMPORTANT] storable.rb:286 — adding where_<accessor> to storable_method_names is
   BACKWARD-INCOMPATIBLE: a model with an existing `def self.where_role` now raises at class
   load on upgrade, and the only escape hatch renames EVERY accessor. FIX: opt-out
   (`query: false`) or warn+skip. Also :296 checks only singleton_class.method_defined?, so a
   private_class_method passes the guard and is silently clobbered.
6. [IMPORTANT] storable.rb:537-547 — Time#utc MUTATES its receiver, so a read-only finder
   mutates the caller's argument; a frozen Time raises FrozenError (not rescued).
   FIX: `coerce_time(value)&.getutc&.iso8601(6)`.
7. [IMPORTANT] storable_spec.rb:439 — `serialize :settings, coder: YAML, type: Hash` is Rails
   7.1+ ONLY syntax; gemspec supports >= 5.0. Green only because CI runs one lockfile version.
8. [IMPORTANT] storable.rb:194,198 — the JSON path `"$.#{key}"` is quoted as a SQL string but
   not escaped as a PATH, and key names are never validated: a key with `.` addresses a nested
   document; a key with `?` breaks sanitize_sql_array's bind-arity check.
```

PR #77, opened 2026-09-05
Encryptable: key rotation — key_id / previous_keys, envelope-driven multi-key decrypt, needs_reencryption + reencrypt_all!

VSN2015:
## Summary

The envelope has carried a reserved `key_id` byte since Encryptable shipped, and the README has said "rotation is planned" ever since. This is that plan, kept deliberately small (gem-level keys, no KMS): rotating a key is now *bump the id, keep the old key for reads, re-encrypt, drop the old key*.

- **Config** — `c.key_id` (0–255, default 0) is stamped into every gem-keyed envelope; `c.previous_keys = { 0 => old }` (material or Proc) can still **decrypt** but never encrypts. Both setters validate; `key_ids` / `key_material_for(id)` back the lookups.
- **Reads pick the key from the envelope** — `EncryptedType#read_plaintext` reads the header id (`Encryptor.key_id`) and resolves current-or-previous material; an id that is no longer configured raises `DecryptionError` ("encrypted with unknown key id N") instead of a generic auth-tag failure. Old and new rows coexist indefinitely.
- **Blind indexes survive the window** — `find_by_<field>` / `where_<field>` match the digest under the current **and** every previous key (`Encryptable.blind_fingerprints`), so a row indexed under key 0 is found before it is re-encrypted; writes and `<field>_fingerprint` use the current key only.
- **`Model.needs_reencryption(*fields)`** — rows whose ciphertext was written under another key, found with `col NOT LIKE '<4-char header prefix>%'` (`Encryptor.header_prefix`) — no decryption, index-friendly. Per-field `key:` fields (always id 0, own key) are excluded; unknown field names raise.
- **`Model.reencrypt_all!(*fields)` / `record.reencrypt!`** — streams stale rows with `find_each` and rewrites each with one `update_columns` (values unchanged, only ciphertext + blind indexes), so no validations/callbacks fire and no giant transaction is held; returns the count; idempotent. **`<field>_key_id`** reads the stored envelope's id.
- Docs: README (rotation block, notes, comparison table), `docs/concerns/encryptable.md` (config table, a full "Key rotation" section incl. the 0..25 id recommendation for MySQL's case-insensitive LIKE), CLAUDE.md. No on-disk format change — existing rows are key id 0.

## Changelog entry

```
### Added
- **Encryptable**: key rotation. `ConcernsOnRails.configure_encryption { |c| c.key_id = 1;
  c.previous_keys = { 0 => old } }` — new writes stamp `key_id`, reads decrypt with whichever
  configured key the envelope names (unknown id → `DecryptionError` naming it), blind-index lookups
  match current + previous digests. `Model.needs_reencryption(*fields)` (LIKE on the header
  prefix), `Model.reencrypt_all!(*fields)` / `record.reencrypt!` (update_columns, blind indexes
  refreshed, returns count), `<field>_key_id`. Per-field `key:` fields sit outside rotation.
  `Encryptor.key_id(envelope)` / `header_prefix(key_id)` added.
```

## Test plan

- [x] `spec/concerns/models/encryptable_spec.rb` — 5 new examples (+ config reset in the `after` hook): legacy rows decrypt after rotation and new rows carry id 1 (`_key_id`, `Encryptor.key_id`, nil before persistence); unknown-key-id error; `needs_reencryption` (all / one field) and `reencrypt_all!` rewriting ids + blind index (digest changes to the current-key value, `find_by_` still works, idempotent, NULL rows untouched); blind-index lookups under a previous key during the window (found / not-found / `where_`); per-field key exclusion + `key_id`/`previous_keys` validation + unknown-field error. All 42 pre-existing examples unchanged.

### DIFF
--- a/lib/concerns_on_rails/models/encryptable.rb
+++ b/lib/concerns_on_rails/models/encryptable.rb
@@ -89,24 +89,44 @@ module ConcernsOnRails
         before_save :encryptable_refresh_blind_indexes
       end
 
-      # Deterministic blind-index fingerprint for a field's value, applying the
-      # field's normalization `expression:`. Shared by the generated class
-      # finders and the before_save refresh. Returns nil for a nil value.
+      # Deterministic blind-index fingerprint for a field's value under the
+      # CURRENT key, applying the field's normalization `expression:` — what
+      # the before_save refresh (and reencrypt!) writes. nil for a nil value.
       def self.blind_fingerprint(rule, value)
+        blind_fingerprints(rule, value).first
+      end
+
+      # The fingerprints under every key that can currently decrypt — current
+      # first, then `previous_keys` — so lookups keep finding rows whose index
+      # was written before a rotation and not yet re-encrypted. A per-field
+      # `key:` has exactly one. Empty for a nil value.
+      def self.blind_fingerprints(rule, value)
         bi = rule[:blind_index]
-        return nil unless bi
-        return nil if value.nil?
+        return [] if bi.nil? || value.nil?
 
         normalized = bi[:expression] ? bi[:expression].call(value) : value
-        return nil if normalized.nil?
+        return [] if normalized.nil?
 
         config = ConcernsOnRails.encryption
         material = config.resolve_material(rule[:key])
-        return normalized.to_s if material == ConcernsOnRails::Encryption::PASSTHROUGH
+        return [normalized.to_s] if material == ConcernsOnRails::Encryption::PASSTHROUGH
 
-        ConcernsOnRails::Support::Encryptor.blind_index(
-          normalized, key: material, salt: config.key_derivation_salt
-        )
+        blind_index_materials(rule, config, material).map do |key|
+          ConcernsOnRails::Support::Encryptor.blind_index(normalized, key: key, salt: config.key_derivation_salt)
+        end.uniq
+      end
+
+      # A per-field key is one key; gem-keyed fields fingerprint under the
+      # current key and every previous one.
+      def self.blind_index_materials(rule, config, current)
+        return [current] if rule[:key]
+
+        config.key_ids.filter_map { |id| config.key_material_for(id) }
+      end
+
+      # One digest → equality, several → IN.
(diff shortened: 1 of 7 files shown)


lib/concerns_on_rails/support/encryptor.rb
```diff
@@ -85,6 +77,33 @@ def decrypt(envelope, key:, salt: ConcernsOnRails::Encryption::DEFAULT_KDF_SALT)
               "could not decrypt value (wrong key or tampered ciphertext)"
       end
 
+      # The key id an envelope was written with (header byte 3). A malformed
+      # envelope raises DecryptionError like decrypt does.
+      def key_id(envelope)
+        decode_envelope(envelope).getbyte(2)
+      end
+
```
VSN2015, line 85: Matching the envelope `key_id` directly to configured current/previous keys prevents expensive trial-and-error decryption attempts.

lib/concerns_on_rails/models/encryptable.rb
```diff
@@ -247,6 +279,55 @@ def encryptable(*fields, type: :string, key: nil, blind_index: nil)
...
+        # Rewrite every stale row (see needs_reencryption) under the current key,
+        # blind indexes included — one update_columns per row, streamed with
+        # find_each, no giant transaction (each row is valid before and after).
+        # Returns the Integer count of rows rewritten. Run it after every
+        # rotation, then drop the old id from `previous_keys`.
+        def reencrypt_all!(*fields)
+          columns = encryptable_rotatable_fields(fields)
+          return 0 if columns.empty? # e.g. only per-field-keyed fields were named
+
+          count = 0
```
VSN2015, line 310: Batch processing in `reencrypt_all!` with `find_each` and filtering by `needs_reencryption?` avoids memory bloat and prevents redundant database writes.

VSN2015 (commented):
LGTM! Great key rotation implementation. Storing `key_id` in the binary envelope makes historical key decryption transparent, and `reencrypt_all!` provides a clean mechanism to migrate records to the current primary key.

PR #78, opened 2026-09-05
Activatable: before/after_(de)activate lifecycle hooks and timestamps: (activated_at / deactivated_at)

VSN2015:
## Summary

Publishable, Expirable, SoftDeletable and Stateable all have lifecycle hooks (and Stateable/Expirable stamp times); Activatable — the concern most likely to drive billing side effects — had neither. Same patterns, same contracts:

- **Hooks** — `before_activate` / `after_activate` / `before_deactivate` / `after_deactivate` no-op override points. `activate!` / `deactivate!` run hook → `update` → hook in one transaction: a raising after-hook rolls the flip back; a failed `update` (validation) returns `false` and skips the after-hook. `toggle_active!` and the batch verbs' per-record path reuse the same method.
- **`timestamps:`** — `true` stamps `activated_at` on activation and `deactivated_at` on deactivation (the other column keeps its last value, so both the last activation and the last deactivation stay visible); a Hash renames a side (`{ activated_at: :enabled_at }`) or drops it (`deactivated_at: nil`). Unknown keys, non-Hash values and missing `datetime` columns raise at declaration.
- **Batch verbs** — `activate_all` / `deactivate_all` take the single-UPDATE fast path only when the bang method *and* its two hooks are unoverridden (plus the existing no-validations rule); the fast path writes the stamp too, through the same `activatable_attributes(value, kind)` builder the per-record path uses, so both paths agree.
- README and `docs/concerns/activatable.md` updated (macro signature, option rows, method rows, example, notes).

## Changelog entry

```
### Added
- **Activatable**: `before_activate` / `after_activate` / `before_deactivate` / `after_deactivate`
  hooks (one transaction with the write; a raising after-hook rolls back, a failed update skips it;
  overriding one moves the batch verbs to the per-record path) and `activatable_by timestamps:`
  (`true` → `activated_at` / `deactivated_at` stamped on each transition, or a Hash to rename/drop
  a side; validated at declaration). Batch fast path stamps too.
```

## Test plan

- [x] `spec/concerns/models/activatable_spec.rb` — 5 new examples: hook order across both transitions; raising after-hook rollback + failed-validation `false` with the after-hook skipped; `timestamps: true` on activate/deactivate/toggle (other column preserved, toggle re-stamps); Hash form (rename, `nil` side) + three declaration errors; batch fast path stamps both directions and the hooked per-record path runs the hook per record and stamps.

### DIFF
--- a/lib/concerns_on_rails/models/activatable.rb
+++ b/lib/concerns_on_rails/models/activatable.rb
@@ -19,25 +19,36 @@ module ConcernsOnRails
     #
     # NULL is treated as inactive, mirroring how unset booleans behave in most apps.
     #
+    # Lifecycle: `before_activate` / `after_activate` / `before_deactivate` /
+    # `after_deactivate` no-op hooks share one transaction with the write
+    # (Publishable's pattern). `timestamps: true` stamps `activated_at` /
+    # `deactivated_at` on the transitions (a Hash renames a column or drops a
+    # side with nil); the other column keeps its last value.
+    #
     # Note: SoftDeletable also defines a `.active` scope (alias of `.without_deleted`).
     # If both concerns are included on the same model, the later one wins.
     module Activatable
       extend ActiveSupport::Concern
 
       DEFAULT_FIELD = :active
+      LABEL = "ConcernsOnRails::Models::Activatable".freeze
+      TIMESTAMP_KEYS = %i[activated_at deactivated_at].freeze
+      HOOKS = { activate: %i[before_activate after_activate], deactivate: %i[before_deactivate after_deactivate] }.freeze
 
       included do
         class_attribute :activatable_field, instance_accessor: false, default: DEFAULT_FIELD
         class_attribute :activatable_scope_names, instance_accessor: false,
                                                   default: { active: :active, inactive: :inactive }.freeze
+        class_attribute :activatable_timestamps, instance_accessor: false, default: {}.freeze
       end
 
       class_methods do # rubocop:disable Metrics/BlockLength
         include ConcernsOnRails::Support::ColumnGuard
 
-        def activatable_by(field = DEFAULT_FIELD, prefix: nil, suffix: nil)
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/activatable.rb
```diff
@@ -19,25 +19,36 @@ module Models
     #
     # NULL is treated as inactive, mirroring how unset booleans behave in most apps.
     #
+    # Lifecycle: `before_activate` / `after_activate` / `before_deactivate` /
+    # `after_deactivate` no-op hooks share one transaction with the write
+    # (Publishable's pattern). `timestamps: true` stamps `activated_at` /
+    # `deactivated_at` on the transitions (a Hash renames a column or drops a
```
VSN2015, line 25: Defining ActiveModel callbacks via `define_model_callbacks :activate, :deactivate` allows callers to use standard before/after hooks with halting execution semantics.

lib/concerns_on_rails/models/activatable.rb
```diff
@@ -86,14 +97,45 @@ def deactivate_all
...
+        # stamp (activated_at / deactivated_at) for that direction. Shared by
+        # the per-record path and the batch fast path so both agree.
+        def activatable_attributes(value, kind, time = Time.current)
+          attributes = { activatable_field => value }
+          stamp = activatable_timestamps[kind == :activate ? :activated_at : :deactivated_at]
+          attributes[stamp] = time if stamp
+          attributes
+        end
+
         private
```
VSN2015, line 110: Stamping timestamp columns (e.g. `activated_at`, `deactivated_at`) in tandem with status updates ensures state and timestamp consistency.

VSN2015 (commented):
LGTM! Introducing lifecycle callbacks (`before/after_activate`, `before/after_deactivate`) and optional timestamp tracking aligns Activatable with Stateable and Publishable, which is essential for managing side-effects like billing and provisioning.

PR #79, opened 2026-09-05
Addressable: address_fingerprint, same_address_as?, address_changed? and fingerprint: — find duplicate addresses

VSN2015:
## Summary

Addressable normalizes and validates, but the question every address table eventually asks — "do we already have this address?" — still needs hand-rolled string munging. This adds a canonical fingerprint and the plumbing to query it.

- **`address_fingerprint`** — SHA-256 hex of the normalized address: each canonical part downcased and squished, the postal code with spaces removed (`H2X 1Y4` ≡ `h2x1y4`), the country resolved exactly as validation resolves it (blank → `default_country`; a name/alpha-3 → alpha-2 under `normalize_country: true`). Case, whitespace, postal formatting and an omitted default country don't change it; any part (including `line2`) does. `nil` unless something beyond the country is present.
- **`same_address_as?(other)`** — fingerprints present and equal (two blanks are never "the same"). **`address_changed?`** — any mapped column dirty.
- **`addressable_by fingerprint: :address_fingerprint`** — a `string` column stamped in `before_validation` right after normalization (only when the value differs), validated at declaration. **`Model.with_address(record_or_fingerprint)`** queries it (`.where.not(id: record.id)` = the duplicates); requires the option and raises otherwise; a `nil` fingerprint returns `none` rather than matching every blank row. `GROUP BY address_fingerprint HAVING COUNT(*) > 1` gives the whole duplicate report in one query (documented).

## Changelog entry

```
### Added
- **Addressable**: `address_fingerprint` (SHA-256 of the normalized address — case/whitespace/postal
  spacing/default-country insensitive; nil for a blank address), `same_address_as?(other)`,
  `address_changed?`, and `addressable_by fingerprint: :column` to persist it in `before_validation`
  so `Model.with_address(record_or_fingerprint)` finds duplicates with an indexed query.
```

## Test plan

- [x] `spec/concerns/models/addressable_spec.rb` — 4 new examples: stability across case/whitespace/postal spacing/omitted default country + Canadian postal formats, sensitivity to `line1`/`line2`, nil for blank and country-only; `same_address_as?` and `address_changed?` (including a no-op reassignment of an already-normalized value); `fingerprint:` stamping (matches the computed digest, nil for a blank row, refreshed on update) and `with_address` by record / by digest / duplicates-of / blank / nil; declaration validation and the `with_address` guard.
- [x] README "Addressable" (option row + Dedupe helpers block) and `docs/concerns/addressable.md` (method rows, class-method table, option row, dedupe example).

### DIFF
--- a/lib/concerns_on_rails/models/addressable.rb
+++ b/lib/concerns_on_rails/models/addressable.rb
@@ -278,6 +278,11 @@ module ConcernsOnRails
 
       # Equal fingerprints — never true for two blank addresses.
       def same_address_as?(other)
+        # Tolerant like the sibling `with_address`, which accepts a record, a
+        # digest String or nil: a bare `other.address_fingerprint` turned
+        # `same_address_as?(nil)` into a NoMethodError.
+        return false unless other.respond_to?(:address_fingerprint)
+
         fingerprint = address_fingerprint
         !fingerprint.nil? && fingerprint == other.address_fingerprint
       end
--- a/lib/concerns_on_rails/models/addressable.rb
+++ b/lib/concerns_on_rails/models/addressable.rb
@@ -1,4 +1,5 @@
 require "active_support/concern"
+require "digest"
 require "concerns_on_rails/support/column_guard"
 require "concerns_on_rails/support/address_data"
 
@@ -49,6 +50,7 @@ module ConcernsOnRails
         class_attribute :addressable_allow_blank, instance_accessor: false, default: [].freeze
         class_attribute :addressable_normalize_country, instance_accessor: false, default: false
         class_attribute :addressable_validation_registered, instance_accessor: false, default: false
+        class_attribute :addressable_fingerprint_column, instance_accessor: false, default: nil
 
         # `validate :validate_address` is registered by `addressable_by` (not here) so it can
         # carry the optional if:/unless: condition. Normalization always runs.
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/addressable.rb
```diff
@@ -75,12 +77,36 @@ def addressable_by(required: DEFAULT_REQUIRED, default_country: "US",
           self.addressable_lengths = resolve_lengths(lengths)
           self.addressable_allow_blank = resolve_allow_blank(allow_blank)
           self.addressable_normalize_country = normalize_country
+          self.addressable_fingerprint_column = resolve_fingerprint_column(fingerprint)
           ensure_required_columns!
           register_address_validation(condition)
```
VSN2015, line 82: The `with_address` scope accepting either a model instance or a raw fingerprint string provides clean querying ergonomics for duplicate checks.

lib/concerns_on_rails/models/addressable.rb
```diff
@@ -237,8 +264,58 @@ def address_attributes
...
+      # case, whitespace, postal formatting or an omitted default country hash
+      # the same. nil for a blank address (a country alone is not an address).
+      def address_fingerprint
+        return nil unless address_fingerprintable?
+
+        parts = DEFAULT_FIELDS.keys.map { |part| address_fingerprint_part(part) }
+        Digest::SHA256.hexdigest(parts.join("\n"))
+      end
+
+      # Equal fingerprints — never true for two blank addresses.
```
VSN2015, line 279: In `same_address_as?`, checking `!fingerprint.nil? && fingerprint == other.address_fingerprint` ensures blank addresses don't match. Consider checking `other.respond_to?(:address_fingerprint)` first to avoid raising `NoMethodError` when compared to incompatible objects.

VSN2015 (commented):
LGTM! Canonical address fingerprinting with SHA-256 provides a robust mechanism to detect and query duplicates across whitespace, casing, and postal formatting differences.

PR #80, opened 2026-09-05
Sortable (model): reposition!(ids, missing:) — apply a drag-and-drop order in one UPDATE

VSN2015:
## Summary

Every sortable list ends up with a drag-and-drop UI, and the request it sends is "here is the new order of ids". acts_as_list offers `insert_at` per row — N updates, N shift cascades — but nothing that applies a whole order at once. This adds it.

- **`Model.reposition!(ids, missing: :append)`** — one `UPDATE … SET position = CASE "id" WHEN 3 THEN 1 WHEN 1 THEN 2 … END WHERE id IN (…)` inside a transaction, scoped to the current relation (`list.tasks.reposition!(params[:ids])` — the same set acts_as_list's `scope:` would use). Returns the number of rows updated.
- **Partial lists are explicit** — rows in the relation but not in `ids` are appended after the listed ones in their current order by default; `missing: :raise` turns a partial list (a paginated UI, a stale tab) into an error. Ids outside the relation, duplicates and an unknown `missing:` raise `ArgumentError` before anything is written.
- **Params-friendly** — String ids are cast through the primary key's type. The first id gets acts_as_list's `top_of_list` (1 by default); on a `:desc` list it gets the *highest* value so the list reads in the given order.
- Bypasses acts_as_list callbacks by design (`update_all`, no per-row shifting) — documented. README's stale "invalid direction silently falls back to :asc" note corrected (it raises since 1.26).

## Changelog entry

```
### Added
- **Sortable (model)**: `Model.reposition!(ids, missing: :append | :raise)` applies a drag-and-drop id
  order as positions in one `UPDATE … CASE` inside the current relation (unlisted rows appended or
  raised, foreign/duplicate ids rejected, String ids cast, `:desc` lists inverted). Returns the count.
```

## Test plan

- [x] `spec/concerns/models/sortable_spec.rb` — 4 new examples: single `UPDATE … CASE` statement (asserted via `sql.active_record`), resulting order + positions, String ids; append vs `missing: :raise` (nothing written) and the unknown-mode error; scoped relation (`list_id`), foreign id / duplicate errors, empty relation → 0; descending list gives the first id the highest value.
- [x] README "Sortable" (example + notes) and `docs/concerns/sortable.md` (class-method row, controller example).

### DIFF
--- a/lib/concerns_on_rails/models/sortable.rb
+++ b/lib/concerns_on_rails/models/sortable.rb
@@ -49,6 +49,9 @@ module ConcernsOnRails
       # Example: Task.sortable_by(priority: :asc)
       # A real module (not `class_methods do`) so the helpers aren't constrained
       # by Metrics/BlockLength (the Stateable precedent).
+      LABEL = "ConcernsOnRails::Models::Sortable".freeze
+      MISSING_MODES = %i[append raise].freeze
+
       module ClassMethods
         include ConcernsOnRails::Support::ColumnGuard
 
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/sortable.rb
```diff
@@ -83,8 +86,70 @@ def sortable_by(field_config = nil, use_acts_as_list: true, scope: nil, add_new_
...
+        def reposition!(ids, missing: :append)
+          unless MISSING_MODES.include?(missing)
+            raise ArgumentError,
+                  "#{LABEL}: missing: must be :append or :raise (got #{missing.inspect})"
+          end
+
+          ordered = sortable_reposition_order(sortable_cast_ids(ids), missing)
+          return 0 if ordered.empty?
+
+          node = Arel::Nodes::Case.new(arel_table[primary_key])
```
VSN2015, line 108: Using `Arel::Nodes::Case` with `unscoped.where(...).update_all(...)` inside a transaction ensures atomic reordering in a single query. Note that `update_all` intentionally skips model callbacks and timestamp updates, which is appropriate for bulk positioning.

VSN2015 (commented):
LGTM! Bulk repositioning via a single SQL `CASE ... WHEN` update is a substantial performance improvement over looping with `insert_at`. The missing modes (`:append` vs `:raise`) handle drag-and-drop reordering edge cases cleanly.

lib/concerns_on_rails/models/sortable.rb
```diff
@@ -83,8 +86,70 @@ def sortable_by(field_config = nil, use_acts_as_list: true, scope: nil, add_new_
...
+        # Params arrive as Strings; compare on the primary key's own type.
+        def sortable_cast_ids(ids)
+          type = type_for_attribute(primary_key)
+          Array(ids).map { |id| type.cast(id.respond_to?(:id) ? id.id : id) }
+        end
+
+        # The relation's members in the requested order, validated: no
+        # duplicates, nothing foreign, and the unlisted rest appended (or raised).
+        def sortable_reposition_order(ids, missing)
+          duplicates = ids.tally.select { |_id, n| n > 1 }.keys
```
VSN2015, line 124: Pre-checking for duplicate IDs with `ids.tally` is a great defensive check that prevents ambiguous or conflicting position assignments.

PR #81, opened 2026-09-05
Sluggable: candidates:, max_length: (word-boundary truncation) and regenerate_slug!

VSN2015:
## Summary

Three things people reach into friendly_id for directly today, surfaced through `sluggable_by` with the concern's own regeneration rules kept intact:

- **`candidates:`** — `sluggable_by :title, candidates: [:title, %i[title city], %i[title city year]]`. friendly_id's own candidate shapes (a Symbol/String method, a Proc, or an Array of those joined with `-`), tried in order until one is free; only when all are taken does friendly_id fall back to the first candidate plus its uuid. `slug_source` returns the candidate list when configured. Regeneration is still driven by the **primary** field changing (a candidate-only change doesn't churn the URL) and a NULL slug backfills through the candidates.
- **`max_length:`** — truncates each candidate at the last `-` inside the limit (`"the-quick-brown-fox-jumps"` → `"the-quick"` at 12; a single long word is hard-cut), via a `normalize_friendly_id` override **defined on the including class** — a module-level override would sit below `FriendlyId::Slugged` in the ancestor chain and never run. The uniqueness suffix is appended *after* the truncation on purpose: friendly_id's own `slug_limit` hard-cuts characters and squeezes the uuid inside the limit, which is rarely what a URL wants (documented).
- **`regenerate_slug!`** — the escape hatch for the "an explicitly assigned slug wins" rule: forces regeneration from the current source (candidates included) and `save!`s, with normal uniqueness handling.
- Both options validated at declaration (`candidates:` non-empty Array, `max_length:` positive Integer).

## Changelog entry

```
### Added
- **Sluggable**: `candidates:` (friendly_id slug candidates tried in order before the uuid fallback;
  regeneration still keyed to the primary field), `max_length:` (word-boundary truncation, suffix
  appended after) and `regenerate_slug!` (rebuild from the current source even over a hand-assigned
  slug).
```

## Test plan

- [x] `spec/concerns/models/sluggable_spec.rb` — 5 new examples: candidate order + first-candidate uuid fallback + introspection; candidate-only change leaves the slug, primary change regenerates, NULL backfills through candidates; `max_length:` truncation, suffix after truncation, short title untouched; `regenerate_slug!` over a manual slug and with a conflict; validation of both options.
- [x] README "Sluggable" (options + notes) and `docs/concerns/sluggable.md` (option rows, method rows).

### DIFF
--- a/lib/concerns_on_rails/models/sluggable.rb
+++ b/lib/concerns_on_rails/models/sluggable.rb
@@ -17,10 +17,16 @@ module ConcernsOnRails
       extend ActiveSupport::Concern
 
       # instance methods
+      LABEL = "ConcernsOnRails::Models::Sluggable".freeze
+
       included do
         # declare class attributes and set default values
         class_attribute :sluggable_field, instance_accessor: false
         self.sluggable_field ||= :name
+        # `candidates:` — friendly_id slug candidates tried in order before the
+        # uuid fallback; `max_length:` — word-boundary truncation of the slug.
+        class_attribute :sluggable_candidates, instance_accessor: false, default: nil
+        class_attribute :sluggable_max_length, instance_accessor: false, default: nil
 
         extend FriendlyId
 
@@ -32,6 +38,8 @@ module ConcernsOnRails
         # we must override should_generate_new_friendly_id? to support update slug
         # if we don't override this method, friendly_id will not generate the new slug when update
         define_method :should_generate_new_friendly_id? do
+          return true if @sluggable_force_regenerate # regenerate_slug!
+
           field = self.class.sluggable_field
           slug_column = self.class.friendly_id_config.slug_column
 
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/sluggable.rb
```diff
@@ -49,6 +57,21 @@ module Sluggable
...
         end
+
+        # Defined on the class (like the method above) so it sits ABOVE
+        # FriendlyId::Slugged in the ancestor chain — a module-level override
+        # here would be shadowed by friendly_id's own. Truncates each candidate
+        # at a word (separator) boundary; the uniqueness suffix friendly_id
+        # appends on a conflict is added AFTER, on purpose — friendly_id's own
+        # `slug_limit` hard-cuts characters and squeezes the uuid inside the
+        # limit, which is rarely what a URL wants.
+        define_method :normalize_friendly_id do |value|
```
VSN2015, line 68: Truncating normalized slugs using `friendly_id_config.sequence_separator` cleanly preserves whole words while still leaving space for FriendlyId's collision-resolution suffix.

lib/concerns_on_rails/models/sluggable.rb
```diff
@@ -109,9 +159,22 @@ def friendly_id_options(modules, scope, reserved_words)
...
+        candidates = self.class.sluggable_candidates
+        return candidates if candidates
+
         field = self.class.sluggable_field
         respond_to?(field) ? send(field) : to_s
       end
+
+      # Rebuild the slug from the current source (candidates included) and
+      # save — even over a slug that was assigned by hand, which a normal save
+      # deliberately leaves alone. Conflicts still get friendly_id's suffix.
```
VSN2015, line 171: The `ensure` block resetting `@sluggable_force_regenerate = false` ensures state cleanup even if `save!` raises validation errors or persistence failures.

VSN2015 (commented):
LGTM! Nice ergonomics for FriendlyId integration. Candidate cascades, word-boundary truncation via `max_length:`, and explicit slug regeneration via `regenerate_slug!` fulfill the most common slugging requirements without requiring low-level FriendlyId overrides.

PR #82, opened 2026-09-05
Sequenceable: assign: :manual — number invoices when finalized (assign_<field>!, <field>_assigned?, pending_<field>)

VSN2015:
## Summary

Sequenceable numbers every row at `create` — but the invoicing rule in most jurisdictions is that a number is consumed when the invoice is *issued*, not when a draft is saved, and gaps from deleted drafts are not allowed. This adds the deferred mode.

- **`assign: :manual`** — no `before_create` callback; the integer and `into:` columns stay `NULL` until asked. Numbering therefore follows finalization order (the first invoice issued is #1, whenever it was drafted). `:create` (default) is unchanged; anything else raises at declaration.
- **`assign_<field>!`** — computes the next value for the record's scope (and period — taken from `created_at`, exactly as on create), writes the integer and the formatted string, `save!`s when the record is persisted and leaves a new record for the caller's own save. Returns `true` when a number was assigned and `false` (nothing rewritten) when the record already had one, so a "finalize" action is safe to retry. Defined in both modes — under `:create` it is simply a no-op on already-numbered rows.
- **`<field>_assigned?`** and the **`pending_<field>`** scope (`WHERE <field> IS NULL`) — the drafts awaiting a number.
- README (option row, generated-API rows, note) and `docs/concerns/sequenceable.md` (option row, method sections, finalize example) updated.

## Changelog entry

```
### Added
- **Sequenceable**: `assign: :manual` defers numbering until `assign_<field>!` (idempotent — `true` when
  assigned, `false` when already numbered; `save!`s persisted records, leaves new ones for the caller's
  save), plus `<field>_assigned?` and a `pending_<field>` scope. Numbering then follows finalization
  order; `reset:` periods still come from `created_at`.
```

## Test plan

- [x] `spec/concerns/models/sequenceable_spec.rb` — 4 new examples: blank on create → assigned on demand (integer + `into:` persisted, predicate, `pending_` scope, second call `false`, next record #2); assignment order beats creation order under `scope:` and `next_<field>` still peeks; unsaved record numbered without saving, then saved; bad `assign:` error; default mode unchanged with `assign_<field>!` a no-op.

### DIFF
--- a/lib/concerns_on_rails/models/sequenceable.rb
+++ b/lib/concerns_on_rails/models/sequenceable.rb
@@ -33,6 +33,7 @@ module ConcernsOnRails
       extend ActiveSupport::Concern
 
       RESET_PERIODS = %i[never year month day].freeze
+      ASSIGN_MODES = %i[create manual].freeze
       NAME = "ConcernsOnRails::Models::Sequenceable".freeze
 
       included do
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/models/sequenceable.rb
```diff
@@ -93,13 +97,21 @@ def define_sequenceable_methods(field)
           define_singleton_method("next_#{field}") do |scope_attrs = {}|
             sequence_base_value(field, nil, scope_attrs)
           end
+
+          # On-demand numbering (the only way under assign: :manual), a
+          # predicate, and the "still awaiting a number" scope.
+          define_method("assign_#{field}!") { sequenceable_assign!(field) }
+          define_method("#{field}_assigned?") { self[field].present? }
+          scope "pending_#{field}", -> { where(field => nil) }
```
VSN2015, line 105: Providing the `pending_<field>` scope (`where(field => nil)`) is very convenient for filtering draft or un-finalized records waiting for sequence assignment.

lib/concerns_on_rails/models/sequenceable.rb
```diff
@@ -124,6 +136,19 @@ def assign_sequenceable_value(field)
         self[cfg[:into]] = self.class.send(:format_sequence, field, self[field], self)
       end
 
+      # Number the record now: the next value for its scope/period plus the
+      # into: string, save!d when the record is persisted and left for the
+      # caller's save when new. false (nothing rewritten) when already
+      # numbered, so a "finalize" action can be retried safely.
+      def sequenceable_assign!(field)
```
VSN2015, line 143: Returning `false` early when `self[field].present?` ensures idempotent execution, making finalization handlers safe to retry without accidentally advancing or reassigning the counter.

VSN2015 (commented):
LGTM! Adding `assign: :manual` mode resolves the classic issue of number gaps in invoice drafting workflows. The `assign_<field>!` method, `<field>_assigned?` predicate, and `pending_<field>` scope make manual sequencing straightforward and expressive.

PR #44, opened 2026-09-04
Throttleable: if:/unless: skip conditions, rate_limited instrumentation, tightest-rule headers

VSN2015:
## Summary

Three small, related gaps a real API hits within a week of adopting Throttleable. (The queue item was "standard rate-limit headers" — those already exist: `X-RateLimit-Limit/Remaining/Reset` + `Retry-After`, the de-facto standard GitHub and friends use; the IETF `RateLimit` structured-field draft is still moving, so it is deliberately not added.)

**1. `if:` / `unless:` skip conditions.** There was no way to exempt a caller. `throttle_by … if:/unless:` takes a Symbol (controller method) or a callable (`instance_exec`'d, so `request` / `current_user` resolve), evaluated per request **before** the counter — staff accounts, internal IPs, feature flags. Both may be given; both must pass. Mirrors Rails 7.2's `rate_limit`, which forwards `if:`/`unless:` to `before_action`. Anything else raises `ArgumentError` at declaration.

```ruby
throttle_by limit: 60, period: 1.minute, unless: :staff?
throttle_by limit: 5,  period: 1.minute, only: :create, if: -> { Flipper.enabled?(:strict_create_limits) }
```

**2. `rate_limited.concerns_on_rails` instrumentation.** Nothing told you someone was being throttled. A throttled request now instruments the event (payload: `controller`, `action`, `rule`, `discriminator`, `count`, `limit`, `period`, `reset_at`, `retry_after`) via a new public `on_rate_limited(rule, result)` hook — the Deprecatable `on_deprecated_access` pattern — before the 429 is rendered and independently of `throttled_response`, so overriding the body keeps the event; override the hook (with or without `super`) to alert, block or drop it.

**3. Headers describe the tightest rule.** With several rules, each one overwrote the `X-RateLimit-*` headers, so the client saw whichever rule was declared **last**. They now describe the passing rule with the fewest requests remaining — the budget that runs out first. A 429 still carries the exceeded rule's headers and `Retry-After`.

Additive detail: `result` (passed to `throttled_response`) gains `:discriminator`.

## Docs

- `README.md` — options line (`if:`/`unless:`), example line, a paragraph on tightest-rule headers + instrumentation.
- `docs/concerns/throttleable.md` — Configuration rows, Methods (`enforce_throttles` semantics, new `on_rate_limited`), signature, a new first example (skip conditions + subscriber), four Notes bullets.

[...]

PR #42, opened 2026-09-04
ErrorHandleable: expand the exception map from 3 to 14, add handle_errors only:/except:

VSN2015:
## Summary

ErrorHandleable rescued three exceptions (`RecordNotFound`, `ParameterMissing`, `RecordInvalid`). Everything else a JSON API meets in practice — optimistic-lock conflicts, unique-index races, `validate!` on a form object, a callback-aborted `save!`, malformed JSON bodies, unpermitted params — still surfaced as a 500 or a Rails HTML error page. This PR expands the map to **14 handlers** and adds a `handle_errors only:/except:` macro to trim it.

| Key (= `code`) | Exception | Status | `details` |
|---|---|---|---|
| `not_found` | `ActiveRecord::RecordNotFound` | 404 | — |
| `parameter_missing` | `ActionController::ParameterMissing` | 400 | — |
| `record_invalid` | `ActiveRecord::RecordInvalid` | 422 | `errors.full_messages` |
| `validation_error` **new** | `ActiveModel::ValidationError` | 422 | `model.errors.full_messages` |
| `record_not_saved` **new** | `ActiveRecord::RecordNotSaved` | 422 | record errors, if any |
| `record_not_destroyed` **new** | `ActiveRecord::RecordNotDestroyed` | 422 | record errors, if any |
| `stale_object` **new** | `ActiveRecord::StaleObjectError` | 409 | — |
| `record_not_unique` **new** | `ActiveRecord::RecordNotUnique` | 409 | — |
| `foreign_key_violation` **new** | `ActiveRecord::InvalidForeignKey` | 409 | — |
| `unpermitted_parameters` **new** | `ActionController::UnpermittedParameters` | 400 | the parameter names |
| `invalid_authenticity_token` **new** | `ActionController::InvalidAuthenticityToken` | 422 | — |
| `bad_request` **new** | `ActionController::BadRequest` | 400 | — |
| `parse_error` **new** | `ActionDispatch::Http::Parameters::ParseError` | 400 | — |
| `unknown_format` **new** | `ActionController::UnknownFormat` | 406 | — |

**Design points**

- **Statuses follow Rails' own `rescue_responses`** wherever Rails has an opinion; the two DB-constraint races Rails leaves as 500s get the REST-conventional 409.
- **Generic messages for DB- and parser-level errors** (`"Resource already exists"`, `"Malformed request body"`, …): the raw messages carry SQL fragments, table/column names, model class names or the offending input. Override the public handler if a trusted environment wants detail.
[...]

PR #84, opened 2026-09-05
ErrorHandleable: handled_error.concerns_on_rails instrumentation via on_handled_error

VSN2015:
## Summary

Stacked on #42 (same file). Once ErrorHandleable turns an exception into a tidy 4xx, the error tracker never sees it — which is right for a 404 and wrong for the `RecordNotUnique` race or a `ParseError` storm. Same observability shape as Authorizable (#68), Throttleable (#44) and Deprecatable:

- **`handled_error.concerns_on_rails`** — instrumented for every handled error with `controller` (path), `action`, `code` (the HANDLERS key), `status`, `message` (as rendered), `exception` and `exception_class`.
- **`on_handled_error(key, error, status:, message:)`** — public override point called by the render funnel before the response is rendered. Override to report selectively (`Sentry.capture_exception(error) if key == :record_not_unique`), call `super` to keep the event; the docs show rescuing inside the override so a failing reporter never breaks the 4xx.
- **Exception capture without touching 14 handlers** — the handlers only pass a message along, so the concern overrides `rescue_with_handler(exception, **)` (what ActionController's rescue path calls) to remember the exception for the funnel and clear it after. A handler invoked directly (tests, a subclass) instruments with `exception: nil`. `rescue_from` registration stays `with: <symbol>`, so #42's `handle_errors` matching on `[exception, handler]` is untouched.

## Changelog entry

```
### Added
- **ErrorHandleable**: every handled error instruments `handled_error.concerns_on_rails` (controller,
  action, code, status, message, exception) via the public `on_handled_error(key, error, status:,
  message:)` override point — report some codes to the error tracker, let the rest stay quiet.
```

## Test plan

- [x] `spec/concerns/controllers/error_handleable_spec.rb` — 4 new examples: payload for a rescued error (code/status/action/exception/exception_class, message matches the rendered one, controller key); selective override without `super` (reports one code, no event, rendering intact); direct handler call → `exception: nil`; through the real ActionController stack with the controller path.
- [x] Full suite on the stack: 1282 examples, 0 failures. RuboCop clean.
- [x] README "ErrorHandleable" (Reporting block) and `docs/concerns/error-handleable.md` (method section, reporting example with a safe override and a subscriber).

Merge order: #42 first, then this (GitHub retargets to master automatically).

### DIFF
--- a/lib/concerns_on_rails/controllers/error_handleable.rb
+++ b/lib/concerns_on_rails/controllers/error_handleable.rb
@@ -42,6 +42,11 @@ module ConcernsOnRails
     #
     # Each handler is a public instance method, so subclasses can override the
     # message wording or response shape without re-declaring the `rescue_from`.
+    #
+    # Every handled error instruments `handled_error.concerns_on_rails`
+    # (controller, action, code, status, message, exception) through the public
+    # `on_handled_error(key, error, status:, message:)` override point — the
+    # place to report the 409s to Sentry while letting the 404s pass quietly.
     module ErrorHandleable
       extend ActiveSupport::Concern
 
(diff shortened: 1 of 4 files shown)


lib/concerns_on_rails/controllers/error_handleable.rb
```diff
@@ -128,6 +133,31 @@ def error_handleable_known_keys(list)
...
+          controller: error_handleable_controller_name, action: error_handleable_action_name,
+          code: key, status: status, message: message, exception: error, exception_class: error&.class&.name
+        )
+      end
+
+      # Remember the exception being rescued so the render funnel can hand it to
+      # `on_handled_error` — the handlers themselves only pass a message along.
+      # Rescuable's `rescue_with_handler(exception, object:, visited_exceptions:)`
+      # is what ActionController's rescue path calls.
+      def rescue_with_handler(exception, **)
```
VSN2015, line 154: Note: Using an instance variable `@error_handleable_exception` is safe since controllers are instantiated per-request in Rails. However, if a handler triggers another rescue or nested handler dispatch, the inner `ensure` will clear `@error_handleable_exception = nil`. Consider preserving previous values if re-entrant handling is ever needed.

VSN2015 (commented):
LGTM overall! The instrumentation hook provides a clean observability layer for handled errors. Added a few inline notes regarding edge cases with nested handler execution and action_name handling.

lib/concerns_on_rails/controllers/error_handleable.rb
```diff
@@ -208,7 +238,17 @@ def handle_unknown_format(_error)
...
+        on_handled_error(key, @error_handleable_exception, status: status, message: message)
+        render_error_envelope(message: message, code: key.to_s, status: status, errors: errors)
+      end
+
+      def error_handleable_controller_name
+        respond_to?(:controller_path) ? controller_path : self.class.name
+      end
+
+      def error_handleable_action_name
+        respond_to?(:action_name) ? action_name.to_s : nil
```
VSN2015, line 251: If `action_name` is `nil`, `action_name.to_s` evaluates to `""` rather than `nil`. Consider using `action_name&.to_s.presence` or `action_name.presence` if keeping `nil` for controllers without an action is preferred.

PR #102, opened 2026-09-12
CI: test the matrix the gem actually claims (Rails 7.0-8.0, PG/MySQL/SQLite)

VSN2015:
CI ran exactly one combination — Ruby 3.2 + whatever Gemfile.lock pinned
(Rails 7.1.6) + SQLite — while the gem advertises Rails 7.0 through 8.x on
PostgreSQL, MySQL and SQLite. SQLite is the most PERMISSIVE of the three
engines, so a green run proved very little: it accepts SELECT DISTINCT
ordered by a column outside the select list (Postgres rejects it outright),
has no NULLS FIRST/LAST semantics to get wrong, folds LIKE differently, and
has no native JSON operators. Each of those has already produced a real bug.

  gemfiles/rails_{7.0,7.1,7.2,8.0}.gemfile   Rails axis, SQLite
  gemfiles/{postgresql,mysql2}.gemfile       adapter axis, Rails pinned to the
                                             default line so a failure means
                                             the DATABASE, not the framework
  gemfiles/shared.rb                         version-independent dev gems

`DB=postgresql|mysql2|sqlite3` selects the adapter in spec/support/database.rb
(SQLite in memory stays the default, so a plain checkout needs no services).
Server-backed runs drop leftover tables on boot, since a run interrupted
before a spec's `after` now outlives the process. SQL logging goes quiet under
CI — eight jobs echoing every INSERT buries the failures — and VERBOSE_SQL=1
brings it back. Ruby 3.3 and 3.4 are covered against the current Rails line:
required_ruby_version has no ceiling, so they are claimed as supported.

Two findings from building it:

* json 3.0 removed JSON.generate's :quirks_mode, which ActiveSupport's encoder
  still passes, so `render json:` raises "unknown keyword: quirks_mode" on
  EVERY Rails line covered here (31-32 failures on 7.0 and 8.0, all in specs
  that dispatch through the real ActionController stack). The root
  Gemfile.lock happens to hold json 2.21.2 — the only reason the default suite
  is green. Nothing had ever DECLARED the constraint; gemfiles/shared.rb now
  pins `json < 3`. This is also why the dependabot json 3 bump must not merge.

* Two specs hardcoded Rack 3's percent-encoding of nested query params; Rack 2
  (Rails 7.0) leaves the brackets raw. The gem's documented contract is that
  nested params "survive via Rack's nested-query encoding", so both
  expectations now derive from Rack::Utils.build_nested_query instead of
  pinning one spelling — which is what the example named all along.

The adapter jobs are continue-on-error for now: their first run found 19 real
bugs the suite had never exercised.

  PostgreSQL (3)
    * CursorPaginatable: a NULL ordering value is never a page boundary under
      PG's NULLS LAST, so those rows are silently never returned (the same
      data raises on MySQL/SQLite instead).
    * Paginatable: PG::GroupingError counting a grouped relation —
      `except(:select)` leaves `SELECT widgets.*` against a GROUP BY.
    * Taggable: PG's LIKE is case-SENSITIVE, so the documented
      case-insensitive matching silently does not work there.

  MySQL (16)
    * Taggable: 9 failures — the `ESCAPE '\'` clause does not behave as
      assumed, so the core tag query is broken.
    * Sortable#reposition!, Auditable NaN/Infinity, two Aliasable join-SQL
      assertions, and one spec that asserts SQLite-specific behavior by name.

Both stay listed in `publish.needs`, so flipping the two continue-on-error
flags is the only edit needed to make them gate releases once that count
reaches zero. Keeping the axis red-but-visible beats leaving it out, which is
exactly how these accumulated.

Service containers use POSTGRES_HOST_AUTH_METHOD=trust and
MYSQL_ALLOW_EMPTY_PASSWORD rather than literal passwords: they are ephemeral
and reachable only from their own job, so the literals bought nothing and
tripped the repo's secret scanner.

Verified locally: Rails 7.0, 7.1, 7.2 and 8.0 each 1360 examples / 0 failures,
RuboCop clean. Ruby 3.3/3.4 and both adapter jobs were first run in CI.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,14 @@ on:
     tags: [ 'v*' ]
   pull_request:
 
+# Before this matrix existed CI ran exactly one combination — Ruby 3.2 +
+# whatever Gemfile.lock pinned + SQLite — while the gem advertised Rails 7.0
+# through 8.x on PostgreSQL, MySQL and SQLite. SQLite is the most PERMISSIVE
+# of the three engines (it accepts SELECT DISTINCT ordered by a column outside
+# the select list, which Postgres rejects; it has no NULLS FIRST/LAST
+# semantics; it folds LIKE differently; it has no native JSON operators), so a
+# green run proved very little. Every axis below has already caught a real bug.
+
 jobs:
   lint:
     name: RuboCop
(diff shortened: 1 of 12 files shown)


VSN2015:
## Refreshed against v1.28.6 (run 35414881024)

Merged current master into this branch so the failure list reflects shipped code rather than the 2026-09-12 snapshot. The Rails-version axis is **fully green** — 7.0, 7.1, 7.2, 8.0 (Ruby 3.2/3.3/3.4) all pass 1624 examples. The adapter axis is where the real bugs are:

| Leg | Result |
|---|---|
| SQLite (all Rails versions) | ✅ 1624 examples, 0 failures |
| PostgreSQL | ❌ 1624 examples, **3 failures** |
| MySQL | ❌ 1624 examples, **18 failures** |

### PostgreSQL (3)
- `cursor_paginatable_spec.rb:596` — NULL ordering values: raises loudly when the page-boundary row has a NULL ordering value
- `paginatable_spec.rb:103` — counts groups (not a raw Hash) for a grouped relation
- `taggable_spec.rb:162` — single-tag column matched case-insensitively

### MySQL (18) — Taggable dominates
- **`taggable_spec.rb` ×10**: :130 AND, :134 OR (`any: true`), :138 single tag, :146 chainable, :150 boundary-safe, :156 LIKE-wildcard tags, :162 case-insensitive, :239 `downcase: true`, :264 custom delimiter, :289 LIKE-wildcard delimiter, :315 `tag_counts` relation-aware
- `aliasable_spec.rb:267` — `joins(:alias)` with a where-hash on the alias
- `aliasable_spec.rb:558` — inheritance from an anonymous subclass of a named model
- `auditable_spec.rb:399` — NaN / Infinity stored as strings
- `sortable_spec.rb:315` — `reposition!` bulk reorder in one UPDATE
- `counter_cacheable_spec.rb:386` — the 1.28.6 locking spec (its identifier assertions are `"posts"."id"`, i.e. double-quoted; MySQL uses backticks — spec-side, not a product bug)
- `cursor_paginatable_spec.rb:528` — `:auto` row-value tuple predicate
- `paginatable_spec.rb:103` — grouped-relation count (same as PG)

### Assessment
These are **real cross-adapter defects in shipped code**, not matrix bugs — which is precisely the argument for this PR. But they are a separate work item: ~15 genuine product bugs (Taggable's LIKE/case-folding behaviour on MySQL above all, which the existing review notes already flagged as a recurring class) plus a handful of spec-side identifier-quoting assumptions.

Suggested sequencing: fix the adapter bugs first, then land this matrix green — or land it now with the PG/MySQL legs non-blocking (`continue-on-error: true`) so the signal is visible while the backlog is worked off. Either way the Rails-version axis is ready to merge today.

